PR #3, opened 2019-12-28
Fixed target name in ask

cizra:


GickerLDS:
Makefile.in is used when running ./configure from the mud root directory, which is needed when setting up the MUD on a new system.

Bobsson:
> Makefile.in is used when running ./configure from the mud root directory, which is needed when setting up the MUD on a new system.

Oops.  We'll probably want to roll back dc5e581799850d142f3b8b3e35bd3f58363fcd62 as well, then.  I'm going to see if I can get Github to just automatically assign pull requests to you, @GickerLDS - you're more familiar with it all than I am.

cizra:
If Makefile.in is used, why is Makefile (an autogenerated file) then committed to repo?
If the configure script is used, why's it not in repo?

GickerLDS:
I see what you're saying, sort of.  You need to manually update Makefile.in with any new files you add to the code, which is then added to Makefile on a new install when doing ./configure.  When working in the repo, you have to manually update both Makefile.in and Makefile with any new files we add.  the configure file is located in the mud directory, and is distributed with those files: the root files, lib, docs, etc.

I don't know if removing the Makefile.in and just using Makefile will break anything, but I'd prefer to leave it be.

cizra:
The fact is that one can compile the project successfully without `Makefile.in` nor `configure`.

As for the list of files — no there's no list of files anywhere around, neither in `Makefile` nor `Makefile.in`. Instead, `Makefile` loops over `*.c` files and autodiscovers them.

GickerLDS:
Ok, I'll take your word for it.  I don't know why it's set up as it is, just that it works.  I've been burned before more than once about people, well-intentioned on cleaning up code, and disappearing with the job half done and me in a mess to fix things up, that's all.

Is this pull request enough, or do I need to re-add your previous one as well?

cizra:
This PR also contains the previous one.
And instead of trusting me, you could check it out and test how well it works :smile: (or whether it works at all, on your machine).

moshehbenavraham:
Thanks Cizra!! @cizra

PR #15, opened 2020-04-28
Support for sending character position via MSDP

rrebrick:


rrebrick:
Any reason this is being held up? I would like to be able to move forward with the MSDPMapper and this will help prevent unnecessary work on the client side to determine character position.

PR #67, opened 2026-08-29
Adversarial spec review: Event-Driven Core Refactor v0.4

moshehbenavraham:
@jamclaug - requesting your review.

Adversarial review of `docs/ongoing-projects/EVENT_DRIVEN_CORE_REFACTOR_SPEC.md` v0.4, checked against both its own text and the Phase 1 scheduler core (`src/game_scheduler.c/.h`, `unittests/CuTest/test_game_scheduler.c`) already on `event-driven-core-refactor`. This PR targets that branch and adds one document plus an index correction. **No source or build change.**

## Headline

The spec is strong and unusually well grounded - every current-state claim in Section 4 that I traced held up. It is not ready for version 1.0. The core problem is a gap between the contract Sections 9-15 publish and what the accepted Phase 1 core actually implements.

## Blocking for 1.0

**F1 - The public contract promises an owner model Phase 1 does not have.** Section 9.1 lists `owner_handle` and owner-index linkage in the event record; Section 10 lists `cancel_owner(owner)`; Section 15.1 requires per-owner capacity; Section 25 requires that owner destruction not scan the global population. `game_scheduler.h` has none of it - no owner kind, no runtime id, no generation, no `cancel_owner`, no owner index, no per-owner limit. Retrofitting this in Phase 5 is not additive: it changes the event record, adds a second index every admission and terminal transition must maintain, adds a failure mode to admission, and changes the cancel surface - after Phases 2-4 have built on the current shape. `cancel_owner` safe-during-dispatch plus O(1) plus cleanup-exactly-once is the single hardest guarantee in the design and the exact thing `mud_event.c` gets wrong today. Recommend pulling the owner index into Phase 1 rather than deferring it.

**F2 - Section 8.3's cascade rule contradicts itself and the code.** Step 2 cascades L1; step 3 then says cascade L2 *before* L1. `advance_one_tick()` at `src/game_scheduler.c:627` implements a third thing - lowest non-zero level first, then slot 0 downward. The code is correct; the spec text is not.

**F3 - Section 7.1 mandates a monotonic clock; nothing in the plan delivers one.** `src/comm.c` paces on `gettimeofday()` at six sites. Phase 2 keeps the heartbeat, Phase 3 preserves timing unchanged, Phase 4 arms a reactor timer, Phase 11 only removes. No phase makes the tick source monotonic. The core itself is clean here - it takes an injected tick source - so this is purely a specification gap.

**F4 - libevent and copyover collide, and the spec never names it.** Copyover is an `execl()` with descriptors deliberately held open, recovered at `src/comm.c:553`. An `event_base` does not survive `exec` - its epoll fd belongs to a dead process image, and `event_reinit()` is for `fork`, not `exec`. Backends may also set `FD_CLOEXEC` or change `O_NONBLOCK` on exactly the descriptors copyover depends on. Separately, the reactor "owning signals" collides with the existing `SIGVTALRM`/`ITIMER_VIRTUAL` checkpointing setup at `src/comm.c:4293`. Phase 3's gate says only the word "copyover".

**F5 - Phase 6's rollback is not a rollback.** The phase deletes pub/sub init, commands, help entries, and rename hooks; the rollback says retain it behind a boot-time boundary. You cannot boot-gate back to something the same phase removed. This is the only phase that breaks Principle 10. Footprint, since Section 20.4 gives no inventory: 17 source files, 3 commands at `src/interpreter.c:3195`, 9 `help.hlp` references, 3 SQL files. Recommend splitting into 6a (introduce, reversible) and 6b (retire, revert-only).

## Also significant

- **F6** - Section 18's thirteen required metrics, Section 10's `inspect(filter)`, and Section 15.1's per-owner limits are assigned to no phase. Shipped `game_scheduler_stats` has 10 fields; no cascade counts, no per-type counts, no overdue age, no admission-failure reasons, and inspect takes a single event id.
- **F7** - Phase 0's defining deliverable, the delay/callback/latency baseline, was never captured. Section 4.8 records build and test validation instead. Every numeric threshold downstream, including reusing 10,000 as the event cap, rests on a measurement that does not exist - and Section 30 then asks reviewers to accept "no unbounded latency regression" with no number attached.
- **F8** - The Section 6 Mermaid graph reuses node id `C` for both the command interpreter and combat encounters. They merge into one node, and the rendered diagram shows the dispatcher feeding the interpreter - the exact architecture Section 6.2 forbids.
- **F9** - Section 16 models Discord and I3 as cross-thread producers, but `discord_bridge`'s sockets are in the select set today (`src/comm.c:1131`, `:1139`). Phase 3 covers listener, descriptors, signals, and cross-thread wakeups - these fall through all four.
- **F10** - Phase ordering puts the reactor rewrite (highest blast radius, per the spec's own risk table) ahead of Phase 5's ownership fix, so a Phase 4 crash has a differential diagnosis spanning new reactor, new wheel, and legacy untyped `void *pStruct` at once.
- **F11** - Section 10 names five errors; the enum has thirteen. `BUSY` (reentrant `advance`), `ID_EXHAUSTED`, and `CLOCK_REVERSED` are all real, reachable, and undocumented, while the `INVALID_OWNER` the spec names does not exist.

Twelve moderate and six minor findings are in the document (Sections 3 and 4), including the missing `Failed` state in the Section 11.1 chart, the unbounded cost of `advance_large` on a full queue, three incompatible statements of the same reentrancy rule, and per-owner-kind generation semantics for rooms under OLC reload.

## Verified clean

Recorded so the next reviewer skips the work: Section 8.1's geometry arithmetic is correct at all five levels and matches the header constants; the 22-test claim is accurate; both build manifests are updated; the core really is inert - `game_scheduler` appears in no other `.c` file; and Section 4's baseline claims check out against source, including the `eCOMBAT_ROUND` structure (`src/combat/fight.c:1855`, `:16557`) and the 10,000 cap at `src/dgscript/dg_event.h:80`.

Section 5 of the review lists what the spec gets right, including the domain-event/decision-hook separation and the tagged result replacing `EVENTFUNC`'s overloaded integer return.

## Disposition

Blocking for 1.0: F1-F5. Before the Phase 1 gate closes: F1, F2, F6, F7, F11. Before Phase 3 starts: F3, F4, F9, F19, F20.

The Phase 1 core is good work and should not be reverted. The gap is closable by amending the spec, moving the owner index into Phase 1, or both.


moshehbenavraham:
Superseded by #85, which consolidates the entire event-driven core refactor line into a single branch. Every commit from this PR is already an ancestor of `refactor/fight-combat-safety`, so nothing is lost by closing this. Closing and deleting the branch to retire the stack.

PR #84, opened 2026-09-02
Refactor combat damage, reactions, and death handling

jamclaug:
## Summary

This PR contains the complete combat-safety and maintainability refactor goal as structured commits.

- harden death handling and reactive defenses, including Life Shield, Greater Hostile Juxtaposition, Divine Sacrifice, and null-killer deaths
- replace recursive reactive damage with a bounded, handle-aware FIFO reaction queue
- introduce typed damage and death outcomes
- remove the duplicate global fighting roster
- validate reaction continuations and report damage dropped at the safety bound
- make combat flow easier to reason about and test

## Deliberate behavior changes

Most of the refactor is behavior-preserving. These four points are not, and are intentional:

1. **Offhand phase-3 attacks now land in phase 3.** The old inline condition tested `PHASE_1`
   where it meant `PHASE_3`, so offhand attacks 3, 6, 9, 12 and 15 ran during phase 1 and phase 3
   got none. `attack_number_runs_in_phase()` implements the intended round-robin
   (`((N - 1) % 3) + 1 == phase`). Characters with many offhand attacks will see fewer hits in
   phase 1 and new hits in phase 3. This is a bug fix, not a rebalance.

2. **`die(ch, NULL)` is no longer a silent no-op.** The old code returned early on a NULL killer,
   so a killer-less death did nothing. Deaths with no attributable killer now resolve through
   `combat_death_apply()` with an explicit cause (`COMBAT_DEATH_SCRIPT` for DG script damage,
   `COMBAT_DEATH_ATTRITION` for quitting while incapacitated). There are no remaining
   `die(x, NULL)` call sites in the tree; every other `die()` caller passes a killer or
   null-checks first.

3. **Reactive damage is deferred, so its `damage()` call returns 0.** While the outermost damage
   call owns an active reaction queue, nested damage (Life Shield reflect, Energy Retort, and
   similar) is scheduled onto that queue and reported as `COMBAT_DAMAGE_QUEUED` — the legacy
   `damage()` return value is 0 and the hit is applied during the drain instead of inline. A
   reflected killing blow therefore resolves slightly later in the round than it used to.
   `combat_state_attack_context_valid()` re-checks both participants after each callback so no
   caller acts on a stale attacker or victim; callers that branched on the return value now see
   "no effect" for that call rather than a death.

4. **Death event ordering is unchanged.** `character_died` still fires at the top of the raw kill
   path, before combat is stopped and before `GET_POS(ch) = POS_DEAD`, so existing listeners see
   the same victim state they always did. The event now carries a `combat_death_cause`.

## Review follow-ups addressed

- Reaction-queue rejections no longer log `SYSERR` for every packet. Only refusal at the safety
  bound is logged, and it now names the dropped damage; packets rejected for a missing
  participant are counted in `queue->stale`.
- `CMakeLists.txt` now lists `combat_damage.h`, `combat_reactions.h` and `combat_state.h`
  alongside `combat_death.h`, matching `Makefile.am`.
- `queue->scheduled` is documented rather than removed: it is the lifetime enqueue total and,
  unlike `count`, does not fall as packets drain, so it is what actually bounds a self-feeding
  reflect chain (A reflects onto B, B reflects back onto A).
- Docstrings added for the new combat contracts and for the functions this diff touches.

## Validation

- CuTest: 1088 tests, all combat/damage/death/reaction suites passing
- Two failures in this worktree (`Test_syntax_check_encounter_world_boots_and_cleans_up_once`,
  `Test_spec_world_binding_source_inventory`) reproduce unchanged on the base branch and are
  environment-related, not caused by this PR
- Earlier full-goal validation: CTest 19/19, ASan/UBSan clean, Valgrind exit status 0


jamclaug:
@coderabbitai review

jamclaug:
Code Review Summary

Verdict: Solid, low-risk refactor with real bug fixes — a few behavior changes worth confirming before merge.

✅ Looks good
- Real crash bug fixed: old Greater Hostile Juxtaposition code did if (af2->modifier <= 0) after a loop that could end with af2 == NULL, and it matched affects by location == SPELL_... (wrong field). The new find_spell_affect() checks affect->spell and null-checks — correct.
- Recursive damage eliminated properly: the FIFO reaction queue (cap 64) is bounded, handle-based (stale targets safely skipped + counted), and drains iteratively. Nested reactions enqueued during the drain get processed — no unbounded recursion.
- combat_list removal is clean: all six combat_list walk-sites converge on combat_state_stop_attackers(), which correctly saves next before stop_fighting(). The initiative-sorted insertion in set_fighting() was dead weight (nothing iterated combat_list for order).
- Null-killer hardening: raw_kill/die now guard every killer deref (GET_MARK, IN_ARENA, info-kill channels) — previously raw_kill(ch, NULL) would NULL-deref.
- Divine Sacrifice now resolves death: paladin reduced to 0 HP by the transferred damage triggers dam_killed_vict — previously the paladin sat at POS_DEAD unprocessed.
- Tests added for every new unit (queue bounds, stale handles, death result, phase mapping); test fixtures correctly migrated off combat_list.

⚠️ Behavior changes to confirm (not blocking, but the PR claims "retain existing gameplay behavior")
1. Offhand phase bug silently fixed — the old inline condition used PHASE_1 where it meant PHASE_3, so phase-3 offhand attacks (3, 6, 9, 12, 15) ran in phase 1 instead of never. The new attack_number_runs_in_phase() is correct, but characters will get fewer offhand hits in phase 1 and new hits in phase 3. Intentional fix? Worth calling out in the PR body.
2. die(ch, NULL) is no longer a no-op — old code returned early with a NULL killer; now it kills. The two known NULL-killer call sites were migrated to combat_death_apply(..., SCRIPT/ATTRITION), but worth grepping for any remaining die(x, NULL) callers.
3. Queued damage returns 0 — nested damage() calls while the reaction queue is active (e.g. Life Shield reflect, Energy Retort) now report "no effect" to their caller and apply later. Callers that branch on the return value (or expect the attacker to be dead immediately after reflect) see different results. The context-validity checks mostly cover this, but delayed reflect-death is a real ordering change.
4. Death event moved later — character_died now fires after GET_POS(ch) = POS_DEAD instead of at the top of raw_kill. Scripts/triggers listening for the event now see a dead victim; verify no script assumes the victim is still "alive-ish."

💡 Minor
- combat_damage_apply logs SYSERR when enqueue rejects invalid handles — an extracted character will spam the log; consider downgrading to a debug-level note or counting it in queue->stale instead.
- CMakeLists.txt lists combat_death.h but not combat_damage.h/combat_reactions.h/combat_state.h (Makefile.am lists all). Headers-only, so no build impact, but inconsistent.
- scheduled counter in the reaction queue is redundant with count given per-call re-init — harmless.

The author's validation claims (19/19 CTest, 1087 CuTests, ASan/UBSan/Valgrind clean) are consistent with the test diff, but worth letting CI confirm.

moshehbenavraham:
Superseded by #85. This branch is now the single consolidated head for the whole event-driven core refactor line, and #85 opens it against `master` with the combat-safety review follow-ups already applied. Closing this PR against the retired `event-driven-core-refactor-test-candidate` base; the branch itself is kept.

Issue #120, opened 2026-09-06
Pets: reconcile schema definitions and strengthen owner/object persistence identities


moshehbenavraham:
The pet incident's atomic snapshot and unconditional versioned migrations were repaired. Remaining design/schema findings are narrower than those resolved failures.

- [ ] Reconcile runtime initialization and `sql/master_schema.sql` field widths, nullability/defaults and unsigned identifiers outside the incident contract; check executable schemas against one migration authority.
- [ ] Replace mutable owner names with stable player IDs through a reviewed migration; enforce the pet-to-object relationship and orphan handling.
- [ ] Audit pet object-load failure paths for recoverable errors instead of terminating the server; verify current behavior before porting any external-game recommendation.
- [ ] Profile complete-owner rewrite churn before choosing dirty-record saves/batching; preserve atomic last-valid snapshots on partial failure.
- [ ] Log query identity/error context without serialized SQL payloads; add meaningful schema, failure-injection and restore checks.

Related #98 covers prepared SQL and #95 portable persistence. This issue retains the specific pet schema/identity and churn gaps. The external sample's missing intelligence insert and weaker runtime-state preservation are not Luminari defects.

Source review: `c7c7d44a7f47e5fc155859eaf359391b827f85ea`. The temporary working notes are being removed; these revision-pinned links retain their evidence and detailed examples. Historical measurements are not fresh runtime results.

- [PET_SYSTEM_COMPARISON_LUMINARI_CHRONICLES_OF_KRYNN.md](https://github.com/LuminariMUD/Luminari-Source/blob/c7c7d44a7f47e5fc155859eaf359391b827f85ea/docs/ongoing-projects/todo-zusuk/PET_SYSTEM_COMPARISON_LUMINARI_CHRONICLES_OF_KRYNN.md)

moshehbenavraham:
Resolved by #146 (merged as ced0bace1). All five checklist items are addressed: one migration authority reconciling the runtime schema and `sql/master_schema.sql`, pet-row identity for saved objects with an enforced cascading foreign key and orphan purge, a written audit confirming pet object loading never terminates the server, a churn measurement that kept whole-owner replacement, and payload-free failure logging with schema, failure-injection, and restore tests. Deliberate non-changes (no `owner_id` / `owner_created` backfill; no dirty-record saves) are documented in `docs/systems/SAVE_SYSTEMS_BREAKDOWN.md`.

PR #147, opened 2026-09-10
Persist temporary summon expiry and define durable follower categories

### LINKED ISSUE #119: Pets: persist temporary summon expiry and define durable follower categories
`save_char_pets()` / `load_char_pets()` preserve substantial runtime state but do not save or recreate the remaining `ePURGEMOB` lifetime. Timed affects and event-based expiry are separate. A temporary follower can reload without its expiration event.

- [ ] Define which follower categories are durable instead of treating every charmed NPC as implicitly persistent.
- [ ] Save a validated lifetime kind and expiry/checkpoint using stable identity; decide offline elapsed-time behavior explicitly.
- [ ] Reject expired followers on load and reconstruct a fresh native event for eligible live ones; never persist a runtime handle.
- [ ] Show remaining expiry/persistence policy in status output and test disconnect, reboot, copyover, repeated saves, malformed records and expired restoration.

Keep the existing versioned runtime-state parser and repaired atomic linked snapshot contract.

Source review: `c7c7d44a7f47e5fc155859eaf359391b827f85ea`. The temporary working notes are being removed; these revision-pinned links retain their evidence and detailed examples. Historical measurements are not fresh runtime results.

- [PET_SYSTEM_COMPARISON_LUMINARI_CHRONICLES_OF_KRYNN.md](https://github.com/LuminariMUD/Luminari-Source/blob/c7c7d44a7f47e5fc155859eaf359391b827f85ea/docs/ongoing-projects/todo-zusuk/PET_SYSTEM_COMPARISON_LUMINARI_CHRONICLES_OF_KRYNN.md)

moshehbenavraham:
Fixes #119

## Problem

`save_char_pets()` / `load_char_pets()` preserved substantial runtime state but not the remaining `ePURGEMOB` lifetime. A temporary follower such as an illusory decoy could reload without its expiration event and live forever. Every charmed NPC was also implicitly persistent, so ordinary spell summons reloaded across logouts.

## Policy

Follower persistence is now explicit (`pet_lifetime_kind()` in `src/utils.c`):

- **Durable** (companions, familiars, mounts, dragon mounts, eidolons, golems, mercenaries, animated dead, lycanthropes, totem spirits, purchased or charmed followers with no spell source): persist until dismissed, killed, or stored.
- **Timed control** (a charm affect with a duration): the remaining affect duration is saved as before and pauses while the owner is offline.
- **Deadline** (a live `ePURGEMOB` event, plus illusory decoys, which never persist without one): an absolute real-time deadline that keeps elapsing offline, across reboots and copyovers.
- **Session** (any other spell summon: summoned creatures, allies, genies, shamblers, planar allies, and so on): lasts while the owner plays and is never saved. A saved record of one is rejected on restore.

The pet keeper boards only durable and timed-control followers. Reclaiming a stored row whose lifetime has ended (for example a decoy stabled before this policy existed) deletes the row and its objects inside the reclaim transaction and tells the player the keeper released it, so the stable slot is not held forever.

## Changes

- Runtime-state record version 4 adds a validated `T <kind> <epoch>` line bound to the existing pet row identity. The scheduler handle is never persisted. The existing versioned parser and the atomic snapshot contract are unchanged.
- Restore rejects an expired record, a decoy record without a deadline, and a session-summon record; the roomless copy is discarded and logged. Active rows are dropped by the next snapshot; stored rows are deleted on reclaim. Live deadline records get a fresh native `ePURGEMOB` for the remaining time, and the restore fails closed if that event cannot be scheduled.
- `save_char_pets()` skips session followers; `pet_store_pet()` and the keeper's `stable store` refuse session and deadline followers with a player-facing message.
- `pets` shows each pet's persistence policy and remaining real time, with a footer explaining offline behavior.
- Help updated in `lib/text/help/help.hlp` and `sql/components/help_pet_entries.sql` (PETS and STABLE). `docs/systems/SAVE_SYSTEMS_BREAKDOWN.md` updated.

## Tests

`unittests/CuTest/test_pet_lifetime.c` covers the record format: deadline round trip with a fresh event, repeated saves agreeing on the absolute deadline, expired restoration rejected with no event, malformed records, missing or stray lifetime lines by version, legacy v3 records restoring as durable, decoys saved as spent without a live event, session summons classified and rejected, keeper acceptance per kind, and status output for every policy.

`Test_pet_lifetime_survives_snapshot_restore_and_keeper_release` in `unittests/CuTest/test_database_persistence.c` drives the production path against MariaDB (`LUMINARI_TEST_MYSQL_ENABLE=1`) with real prototypes and a room: `save_char_pets()` skips the session summon and stores the absolute deadline; `load_char_pets()` republishes the deadline follower with a fresh event and the control follower with its affect; an expired active row is refused on reload while the roster still counts as fully restored and the next `save_char_pets()` drops it; `pet_store_pet()` refuses deadline and session followers; `pet_retrieve_stored()` deletes a spent legacy stored decoy and reports the release, and still reclaims a live durable stored follower into the room. Disconnect, reboot, and copyover all reach persistence through these same functions.

https://claude.ai/code/session_01Adu1bF5vNR7Drp3a9oKR97

### DIFF
--- a/docs/systems/SAVE_SYSTEMS_BREAKDOWN.md
+++ b/docs/systems/SAVE_SYSTEMS_BREAKDOWN.md
@@ -196,12 +196,26 @@ repeated renaming does not accumulate prior custom names. Saved eidolon identity
 is restored before considering legacy owner-description defaults. Malformed
 records remain saved for recovery; `pets restore` offers a bounded retry and skips already published pet IDs.
 
-Timed affects currently retain remaining duration but pause while stored/offline.
-The runtime-state record does not preserve natural/control event deadlines.
-`ePURGEMOB` is the short follower-loss cleanup, not a general natural lifetime.
-Absolute deadlines, expiry gear handling, and uncertain-commit reconciliation
-remain open in `docs/ongoing-projects/PET_SYSTEM_REFACTOR_PLAN.md`. Save/load tests
-do not replace the plan's executable copyover acceptance gate.
+Follower persistence follows an explicit policy (`pet_lifetime_kind()` in
+`src/utils.c`). Durable followers persist until dismissed, killed, or stored.
+Timed control is carried by the saved charm affect; its remaining duration
+pauses while stored or offline. Deadline followers are those with a live
+`ePURGEMOB` event, plus illusory decoys, which never persist without one. The
+runtime-state record (version 4, `T <kind> <epoch>`) stores the absolute
+real-time deadline rather than the scheduler handle. The deadline keeps
+elapsing offline: restore rejects an expired record, discards the roomless
+copy, and lets the next snapshot drop the row; a live record gets a fresh
+native `ePURGEMOB` event for the remaining time; if that event cannot be
+scheduled the record is rejected rather than restored without an expiry.
+Session followers are spell summons outside the kept families (companion,
+familiar, mount, dragon mount, eidolon, golem, mercenary, animated dead,
+lycanthrope, totem spirit); they are skipped by the snapshot and a saved record
+of one is rejected on restore. The keeper boards only durable and timed-control
+followers; reclaiming a stored row whose lifetime has ended deletes that row
+and its objects inside the reclaim transaction. `pets` shows each pet's
+policy and remaining real time. Expiry gear handling and uncertain-commit
+reconciliation remain open in `docs/ongoing-projects/PET_SYSTEM_REFACTOR_PLAN.md`.
+Save/load tests do not replace the plan's executable copyover acceptance gate.
 
 #### 3. Wilderness System Data
 - **Tables**: `region_data`, `path_data`, `region_index`, `path_index`
--- a/src/players.c
+++ b/src/players.c
@@ -63,7 +63,7 @@
 #define PLAYER_AFFECT_FILE_VERSION 1
 #define BOARDING_ABILITY_PFILE_VERSION 1
 
-#define PET_RUNTIME_STATE_VERSION 3
+#define PET_RUNTIME_STATE_VERSION 4
 #define PET_RUNTIME_STATE_INITIAL_SIZE 4096
 #define PET_RUNTIME_STATE_MAX_SIZE 65535
 
@@ -181,7 +181,14 @@ struct pet_runtime_state
   bool mercenary_proc_fired;
   int source_spell;
   int behavior;
+  int lifetime_kind;    /* PET_SAVED_LIFETIME_* */
+  long long expires_at; /* Real-time epoch for deadline records; zero otherwise. */
 };
+
+/* Persisted lifetime kinds.  Timed control affects are already carried by
+ * the affect records, so only the event deadline needs its own marker. */
+#define PET_SAVED_LIFETIME_DURABLE 0
+#define PET_SAVED_LIFETIME_DEADLINE 1
 
 static char *serialize_pet_runtime_state(struct char_data *pet);
 static bool parse_pet_runtime_state(const char *serialized, struct pet_runtime_state *state);
@@ -5939,6 +5946,16 @@ static char *serialize_pet_runtime_state(struct char_data *pet)
   state.mercenary_proc_fired = state.hired_mercenary && PROC_FIRED(pet);
   state.source_spell = pet->pet_source_spell;
   state.behavior = pet->pet_behavior;
+  /* The scheduler handle is runtime-only; persist the absolute deadline it
+   * represents.  A deadline-class follower without a live event is already
+   * spent, so it is saved as expired rather than promoted to durable. */
+  if (pet_lifetime_kind(pet) == PET_LIFETIME_DEADLINE)
+  {
+    state.lifetime_kind = PET_SAVED_LIFETIME_DEADLINE;
+    state.expires_at = pet_lifetime_deadline(pet);
+    if (state.expires_at <= 0)
+      state.expires_at = (long long)time(NULL);
+  }
   for (i = 0; i < NUM_OF_SAVING_THROWS; i++)
     state.saves[i] = GET_REAL_SAVE(pet, i);
   for (i = 0; i < 10; i++)
@@ -5962,11 +5979,13 @@ static char *serialize_pet_runtime_state(struct char_data *pet)
       goto serialize_failure;                                                                      \
   } while (0)
 
-  /* V=version, B=affect bits, M=mobile bits, S=stats, R=saves, L=spell slots,
+  /* V=version, P=source spell, H=behavior, T=lifetime kind and deadline,
+   * B=affect bits, M=mobile bits, S=stats, R=saves, L=spell slots,
    * F=feat override, A=timed affect, and E=end. */
   PET_STATE_APPEND("V %d\n", PET_RUNTIME_STATE_VERSION);
   PET_STATE_APPEND("P %d\n", state.source_spell);
   PET_STATE_APPEND("H %d\n", state.behavior);
+  PET_STATE_APPEND("T %d %lld\n", state.lifetime_kind, state.expires_at);
   PET_STATE_APPEND("B %d %d %d %d %d %d %d %d\n", state.extra_aff[0], state.extra_aff[1],
                    state.extra_aff[2], state.extra_aff[3], state.extra_aff2[0], state.extra_aff2[1],
                    state.extra_aff2[2], state.extra_aff2[3]);
@@ -6043,6 +6062,10 @@ static bool pet_state_values_are_valid(const struct pet_runtime_state *state)
       state->source_spell < 0 || state->source_spell >= MAX_SPELLS || state->behavior < 0 ||
       state->behavior >= NUM_PET_BEHAVIORS)
     return false;
+  if (state->lifetime_kind == PET_SAVED_LIFETIME_DURABLE
+          ? state->expires_at != 0
+          : state->lifetime_kind != PET_SAVED_LIFETIME_DEADLINE || state->expires_at <= 0)
+    return false;
 
   for (i = 0; i < NUM_OF_SAVING_THROWS; i++)
     if (state->saves[i] < SHRT_MIN || state->saves[i] > SHRT_MAX)
@@ -6073,6 +6096,7 @@ static bool parse_pet_runtime_state(const char *serialized, struct pet_runtime_s
   int consumed, feat, feat_value, hired, proc_fired, version;
   int saw_version, saw_base, saw_mob, saw_stats, saw_saves, saw_slots, saw_end, saw_source;
   bool saw_behavior;
+  bool saw_lifetime;
   int affect_values[14];
   int values[20];
 
@@ -6090,6 +6114,7 @@ static bool parse_pet_runtime_state(const char *serialized, struct pet_runtime_s
   saw_version = saw_base = saw_mob = saw_stats = saw_saves = saw_slots = saw_end = false;
   saw_source = false;
   saw_behavior = false;
+  saw_lifetime = false;
   version = 0;
 
   for (line = strtok_r(copy, "\n", &saveptr); line; line = strtok_r(NULL, "\n", &saveptr))
@@ -6121,6 +6146,14 @@ static bool parse_pet_runtime_state(const char *serialized, struct pet_runtime_s
           !pet_state_line_is_complete(line, consumed))
         goto parse_failure;
       saw_behavior = true;
+    }
+    else if (line[0] == 'T')
+    {
+      if (!saw_version || version < 4 || saw_lifetime ||
+          sscanf(line, "T %d %lld %n", &state->lifetime_kind, &state->expires_at, &consumed) != 2 ||
+          !pet_state_line_is_complete(line, consumed))
+        goto parse_failure;
+      saw_lifetime = true;
     }
     else if (line[0] == 'B')
     {
@@ -6237,7 +6270,7 @@ static bool parse_pet_runtime_state(const char *serialized, struct pet_runtime_s
   free(copy);
   if (!saw_version || !saw_base || !saw_mob || !saw_stats || !saw_saves || !saw_slots || !saw_end ||
       (version >= 2 && !saw_source) || (version >= 3 && !saw_behavior) ||
-      !pet_state_values_are_valid(state))
+      (version >= 4 && !saw_lifetime) || !pet_state_values_are_valid(state))
     return false;
 
   mask_pet_state_bits(state->extra_aff, AF_ARRAY_MAX, NUM_AFF_FLAGS);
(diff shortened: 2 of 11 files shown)


src/players.c
```diff
@@ -7189,6 +7244,14 @@ static struct char_data *prepare_saved_pet_row(struct char_data *ch, MYSQL_ROW r
     extract_char(mob);
     return NULL;
   }
+  /* The next snapshot drops the expired row; nothing else about the roster failed. */
+  if (!restore_pet_lifetime(mob, has_runtime_state ? &runtime_state : NULL, time(NULL)))
```
moshehbenavraham, line 7268: **[blocker]** Expired restore is a non-failure extract of a roomless copy. That only lets the **next active snapshot** drop the row.

`pet_retrieve_stored` (not in this diff, ~7818) shares this decoder. On NULL it sets "The keeper cannot rouse that follower." and ROLLBACK, so `PET_STATE_STORED` is unchanged. `save_char_pets` only DELETEs/replaces `PET_STATE_ACTIVE` rows.

**Claimed:** reject expired, discard roomless copy, next snapshot drops the row.

**Actual:** stored deadline pets (manual `stable store`, or `pet_store_surviving_followers` on death, both of which accept decoys) stay in the stable forever after the epoch, occupying one of 10 slots. Help says the pet is gone when you return.

Fix on this path: if lifetime restore fails because the deadline elapsed (or a decoy had no deadline), delete or mark-unretrievable the **stored** row and its `pet_save_objs` instead of rolling back to the expired snapshot. Do not set `restore_failed` for that case on login (active drop-on-snapshot is correct).

src/players.c
```diff
@@ -6309,6 +6342,28 @@ static void apply_pet_runtime_state(struct char_data *pet, const struct pet_runt
...
+  if (!pet)
+    return false;
+  if (!state || state->lifetime_kind == PET_SAVED_LIFETIME_DURABLE)
+    return !pet_lifetime_requires_deadline(pet);
+  remaining = state->expires_at - (long long)now;
+  if (remaining <= 0)
+    return false;
+  if (remaining > LONG_MAX / PASSES_PER_SEC)
+    remaining = LONG_MAX / PASSES_PER_SEC;
+  attach_mud_event(new_mud_event(ePURGEMOB, pet, NULL), (long)remaining * PASSES_PER_SEC);
```
moshehbenavraham, line 6368: **[non-blocker]** `attach_mud_event` is void; on admission failure it frees the event and returns. This function still `return true` and the caller publishes the pet.

**Claimed:** a live record gets a fresh native `ePURGEMOB` for the remaining time.

**Actual:** the call is issued; success is not observed. A deadline follower restored without an event is the original #119 bug for that session (a decoy would be saved spent on the next snapshot, so it is not forever).

After attach, require `mud_event_is_live(char_has_mud_event(pet, ePURGEMOB))` and return false (discard) if it is not live.

unittests/CuTest/test_pet_lifetime.c
```diff
@@ -0,0 +1,283 @@
...
+
+  CuAssertIntEquals(tc, PET_LIFETIME_DEADLINE, source_kind);
+  /* The saved deadline is absolute real time, so repeated saves agree. */
+  CuAssertTrue(tc, first_deadline >= now + 89 && first_deadline <= now + 92);
+  CuAssertTrue(tc, second_deadline >= first_deadline && second_deadline <= first_deadline + 1);
+  CuAssertTrue(tc, restored_ok);
+  CuAssertTrue(tc, remaining >= 87L * PASSES_PER_SEC && remaining <= 92L * PASSES_PER_SEC);
+}
+
+void Test_pet_lifetime_rejects_expired_and_malformed_records(CuTest *tc)
```
moshehbenavraham, line 139: **[non-blocker]** This suite drives `serialize_pet_runtime_state` / `restore_pet_runtime_state_for_test` only. Expired is "helper returned false and no event," not "`prepare_saved_pet_row` extracted a roomless `read_mobile` copy and the DB row was dropped."

Missing vs the PR/#119 claims:

- `save_char_pets` / `load_char_pets` (disconnect, reboot, copyover all go through those)
- keeper retrieve of an expired `T 1 <past>` stored row (`pet_retrieve_stored` rollback vs active snapshot drop)
- `restore_failed` left false on expiry so login can still mark `PET_ROSTER_LOADED`

The serializer cases here (malformed `T` lines, v3 stray `T`, spent decoy `T 1 <now>`) are real. They would not have caught the stored-row leak.

docs/systems/SAVE_SYSTEMS_BREAKDOWN.md
```diff
@@ -157,12 +157,19 @@ repeated renaming does not accumulate prior custom names. Saved eidolon identity
...
-do not replace the plan's executable copyover acceptance gate.
+Follower persistence follows an explicit policy (`pet_lifetime_kind()` in
+`src/utils.c`). Durable followers persist until dismissed, killed, or stored.
+Timed control is carried by the saved charm affect; its remaining duration
+pauses while stored or offline. Deadline followers are those with a live
+`ePURGEMOB` event, plus illusory decoys, which never persist without one. The
+runtime-state record (version 4, `T <kind> <epoch>`) stores the absolute
+real-time deadline rather than the scheduler handle. The deadline keeps
+elapsing offline: restore rejects an expired record, discards the roomless
+copy, and lets the next snapshot drop the row; a live record gets a fresh
```
moshehbenavraham, line 207: **[non-blocker / docs]** "lets the next snapshot drop the row" is only true for `PET_STATE_ACTIVE` after `load_char_pets` succeeds. Stored rows have no such snapshot: `save_char_pets` replaces active rows only, and keeper retrieve rolls back on expiry.

Same overclaim is in the PR body and the `prepare_saved_pet_row` comment at `src/players.c:7247`. If stored deadline pets are meant to expire, document that reclaim deletes the stored row; if they are meant to pause, `pet_lifetime_kind`/serialize must not keep a wall-clock epoch while `PET_STATE_STORED`.

moshehbenavraham (commented):
## Verdict: issues found

Adversarial review of head `b8481b80e17f3c12009f0bbf1b5c599f77321930` (all 10 files plus `save_char_pets` / `load_char_pets` / `prepare_saved_pet_row` / `pet_retrieve_stored` / `ePURGEMOB` / charm-duration / copyover login). CodeRabbit's rate-limit stub is not this review. `#120` / `pet_save_objs.pet_idnum` is out of this diff.

The version-4 `T <kind> <epoch>` parser, spent-decoy serialization (`T 1 <now>` rather than durable), and **active** roster login path are largely what #119 asked for: the scheduler handle is never persisted, malformed blobs still fail parse and retain the row, and an expired active pet is extracted roomless without setting `restore_failed` so a later `save_char_pets` snapshot can drop it. Copyover really does `save_player_pets()` then `enter_player_game()` -> `load_char_pets()`.

The hole is the shared decoder's expiry path vs **stored** rows.

### Blocker

1. **Expired stored deadline pets are never dropped** (`src/players.c:7248`, caller `pet_retrieve_stored` ~7818 is not in the diff).
   - **Claimed:** restore rejects an expired record, discards the roomless copy, and the next snapshot drops the row.
   - **Actual:** true only for `PET_STATE_ACTIVE` after a successful load plus a later snapshot. `pet_retrieve_stored` treats `prepare_saved_pet_row` returning NULL as "The keeper cannot rouse that follower." and **ROLLBACK**s, so `pet_state` stays `PET_STATE_STORED`. `save_char_pets` only replaces active rows.
   - **Why it matters:** `PET_KEEPER_CAPACITY` is 10. `stable store` accepts any loyal follower, including an illusory decoy. `pet_store_surviving_followers()` auto-stores every charmed NPC on death, including decoys. After the 120s `SPELL_MISLEAD` `ePURGEMOB` elapses, reclaim fails forever and the slot stays occupied. Help says the pet is gone when you return; the stable list still shows it.

### Non-blockers

2. **`restore_pet_lifetime` does not verify the fresh `ePURGEMOB` attached** (`src/players.c:6363`). `attach_mud_event` is void and frees the event on admission failure; the function still `return true` and the pet is published. That is session-length immortality on the reconstruct path (a later decoy save would write spent `T 1 <now>`). Claimed: live records get a fresh native event.

3. **New tests never drive the production restore/snapshot/keeper path** (`unittests/CuTest/test_pet_lifetime.c`). They exercise `serialize_pet_runtime_state` / `restore_pet_runtime_state_for_test` only (deadline round-trip, repeated serialize, malformed `T` lines, v3 stray line, spent decoy, PETS strings). They do not call `save_char_pets`, `load_char_pets`, `prepare_saved_pet_row`, or `pet_retrieve_stored`. Issue #119's disconnect/reboot/copyover coverage is call-graph commentary. Finding 1 is exactly the untested path.

4. **Save eligibility is still "every charmed NPC"** (`pet_lifetime_kind` / `save_char_pets`). #119 asked to define durable categories instead of treating every charmed NPC as implicitly persistent. Magical summons still set the charm bit without `ePURGEMOB` or a charm affect with `duration >= 0` (only `SPELL_MISLEAD` gets the 120s event), so they classify as `PET_LIFETIME_DURABLE` and still reload forever. Labels changed; what gets snapshotted did not.

### Checked and not wrong

- Durable serialize is `T 0 0` (`init_pet_runtime_state` memset), not stack garbage.
- Parser `%n` + `pet_state_line_is_complete` rejects trailing junk; v4 requires `T`; v3 stray `T` fails; `T 0 5` / `T 1 0` / `T 2 5` fail validation.
- Absolute epoch + pulse ceiling: repeated saves stay within ~1s; not an immortality leak.
- NPC `ePURGEMOB` is not in the player event copyover table; pets rebuild from DB only (no double event).
- `stop_follower`'s 12s purge is after the pet leaves the follower list, so it is not snapshotted as a deadline companion.
- Help `.hlp` and `help_pet_entries.sql` match aside from SQL `pet''s` escaping.

moshehbenavraham:
Addressed all four review findings in the follow-up commit (branch squashed to one commit on top of current master; no rebase was needed).

**1. Expired stored deadline pets were never dropped (blocker).** `prepare_saved_pet_row()` now reports lifetime rejection through an `expired` out-parameter, distinct from decode failure. `pet_retrieve_stored()` deletes the spent row and its `pet_save_objs` inside the already-open `FOR UPDATE` transaction, commits, and tells the player the keeper released it, so the slot is freed. Separately, `pet_store_pet()` and the keeper's `stable store` now refuse deadline and session followers (`pet_keeper_accepts()`), so new decoys cannot be stabled at all; `pet_store_surviving_followers()` inherits that through `pet_store_pet()`. Active rows still drop on the next snapshot as before.

**2. `restore_pet_lifetime()` did not verify the fresh `ePURGEMOB` attached.** After `attach_mud_event()` it now checks `mud_event_is_live(char_has_mud_event(pet, ePURGEMOB))` and fails closed with a SYSERR, so an admission failure rejects the record instead of publishing an immortal follower.

**3. Tests never drove the production restore/snapshot/keeper path.** Added `Test_pet_lifetime_survives_snapshot_restore_and_keeper_release` in `test_database_persistence.c` (runs with `LUMINARI_TEST_MYSQL_ENABLE=1`, real prototypes and a room): `save_char_pets()` stores the absolute deadline and skips the session summon; `load_char_pets()` republishes the deadline follower with a fresh live event and the control follower with its affect; an expired active row is refused on reload while the roster still counts as loaded, and the next `save_char_pets()` drops it; `pet_store_pet()` refuses deadline and session followers; `pet_retrieve_stored()` deletes a spent legacy stored decoy with the release reason and still reclaims a live durable stored follower into the room. Finding 1's exact path is covered.

**4. Save eligibility was still every charmed NPC.** Added `PET_LIFETIME_SESSION`: a follower with a spell source (`pet_source_spell != 0`) that is not in a kept family (companion, familiar, mount, dragon mount, eidolon, golem, mercenary, animated dead, lycanthrope, totem spirit) and has neither a deadline nor timed control. `save_char_pets()` skips session followers, `restore_pet_lifetime()` rejects a saved record of one (including legacy rows), the keeper refuses them, and `pets` labels them "this session only, not saved". Animated dead stay durable because HELP ANIMATE DEAD promises a permanent follower. The kept-family list is one table in `src/utils.c` if the policy needs adjusting.

Also fixed an `int` narrowing in `pet_lifetime_deadline()` flagged by clang-tidy, and updated HELP PETS, HELP STABLE (both `help.hlp` and `sql/components/help_pet_entries.sql`, applied to the dev help database) and `docs/systems/SAVE_SYSTEMS_BREAKDOWN.md`.

**Local CI replication** (GitHub jobs are not treated as gates):
- `make test` + `make install`: 1367 runs, 1366 pass. The single failure is `Test_syntax_check_encounter_world_boots_and_cleans_up_once`, which forks the server against the local dev database that the in-progress #120 branch has already migrated (`SYSERR: Required pet column pet_save_objs.pet_idnum has an incompatible type or nullability`); the same reason the local server startup smoke test cannot run here. Not related to this diff, and it passes on CI's isolated database.
- All remaining `make test` sub-targets, `make test-world-tools`, `make test-protocol`, protocol fuzzing (15 s), `make test-process-memory`, `test-character-rename-static/schema`: pass.
- Valgrind (`--errors-for-leak-kinds=definite`) over `cutest`: 0 errors apart from CuTest's own failure-message strings for the environment failure above.
- ASan/UBSan `cutest` build (gcc, same flags as CI): no sanitizer errors. With the MariaDB tests enabled (CI's sanitizer job does not enable them) LeakSanitizer reports 11 small leaks in the pre-existing `Test_named_eidolon_keeper_restore_preserves_saved_identity` path (`set_eidolon_descs` strdup via `pet_retrieve_stored`); identical at the previous PR head, so pre-existing and out of scope. Worth a follow-up issue.
- Strict-warning rebuild of changed files (`-Wall -Wextra`): 0 warnings. `pre-commit run clang-format`: clean. clang-tidy (CI check set) on `src/utils.c`: no findings in the changed code. gitleaks: no leaks.

moshehbenavraham:
Merged to master as 303ba07c4 (squash of one commit, rebased onto #146 first with no conflicts).

**What merged**
- Explicit follower persistence policy (`pet_lifetime_kind()`): durable, timed control, real-time deadline, and session-bound spell summons. Session summons are never snapshotted and their saved records are rejected on restore.
- Runtime-state record version 4 with a validated `T <kind> <epoch>` deadline line; restored deadlines rebuild a native `ePURGEMOB` event and fail closed if it cannot be scheduled.
- Keeper boards only durable and timed-control followers; reclaiming a stored row whose lifetime has ended deletes it inside the reclaim transaction and releases the slot.
- `pets` shows each pet's policy and remaining real time. HELP PETS and HELP STABLE updated in `help.hlp` and the SQL component (applied to the dev help database); `docs/systems/SAVE_SYSTEMS_BREAKDOWN.md` updated.
- Unit coverage of the record format plus a MariaDB-backed test through `save_char_pets`, `load_char_pets`, `pet_store_pet`, and `pet_retrieve_stored`.

Rebased verification: `make test` on top of #146 ran 1370 tests with 1369 passing; the single failure is the syntax-check boot test against the local dev database, which now trips on `pet_data` description-column drift from #146's schema contract, not on this change.

**Follow-up**
- #148: pre-existing LeakSanitizer findings in the eidolon keeper round-trip test path (only visible with the MariaDB tests enabled under ASan).
- Expiry gear handling and uncertain-commit reconciliation remain open in `docs/ongoing-projects/PET_SYSTEM_REFACTOR_PLAN.md`, as noted in the save-systems document.
- The local dev database needs re-aligning with `sql/master_schema.sql` before the server or the syntax-check boot test will run here.

PR #153, opened 2026-09-11
Own a real optimized and hardened production build profile

### LINKED ISSUE #93: Modernization: own a real optimized and hardened production build profile
## Why

Production deployment requests an optimization mode that does not exist, while observed binary hardening currently depends largely on Ubuntu toolchain defaults rather than a repository-owned contract.

- `scripts/deployment/deploy.sh --prod` invokes `./configure --enable-optimizations`.
- `configure.ac` defines no matching `AC_ARG_ENABLE`; configure reports `unrecognized options: --enable-optimizations`.
- The resulting flags remain the configure default (`-g -O2` in the audited environment), so the deployment interface is misleading.
- The audited Ubuntu-built binary happened to be PIE, use a non-executable stack, RELRO/NOW, stack protector, and fortified libc calls, but the project does not request or verify those properties.
- Autotools and CMake do not expose equivalent hardened Release profiles.

## Scope

- Replace the no-op option with a real, documented production profile in both Autotools and CMake, and make deployment fail on unknown configure options.
- Define optimization policy (`-O2` or measured alternative), debug-symbol retention/splitting, assertion policy, and crash-symbolization workflow.
- Feature-detect and enable a coherent hardening set. For current GCC, prefer `-fhardened` where supported; otherwise apply documented equivalents such as:
  - `_FORTIFY_SOURCE=3` at optimization;
  - PIE;
  - full RELRO and immediate binding;
  - strong stack protector;
  - stack-clash protection;
  - control-flow protection where architecture/toolchain support it;
  - non-executable stack.
- Define the equivalent supported Clang configuration without passing unsupported flags blindly.
- Make optional LTO/ThinLTO and PGO explicit measured profiles, not default assumptions.
- Verify final ELF properties and dangerous linker/runtime dependencies with `readelf`, `checksec`, `annocheck`, or an equivalent reproducible check.
- Add a small performance/latency benchmark suite for the game loop and hot parsers so hardening/optimization choices are evidence based.
- Package or retain build IDs and debug symbols appropriate for production crash analysis.

## Acceptance criteria

- `deploy.sh --prod` selects a real profile and no longer emits an unrecognized-option warning.
- Autotools and CMake produce equivalently optimized/hardened binaries for supported GCC and Clang.
- A CI job inspects the linked binary and fails when required properties regress.
- Unsupported hardening features are detected and reported, not silently ignored.
- Release tests run against the exact hardened artifact.
- Debugging and symbol retention remain possible for production crashes.
- Performance impact is recorded against a stable benchmark before making LTO/PGO default.

## Reference

- GCC instrumentation and `-fhardened`: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html

moshehbenavraham:
Fixes #93.

## Why

`deploy.sh --prod` invoked `./configure --enable-optimizations`, an option `configure.ac` never defined. Production builds silently fell back to the default `-g -O2` flags and every observed hardening property came from Ubuntu toolchain defaults rather than a repository-owned contract.

## What changed

- **One shared profile contract.** `scripts/deployment/production_profile.sh` feature-detects the flags for the selected compiler and is called by both `configure.ac` (`--enable-production`) and `CMakeLists.txt` (`LUMINARI_PRODUCTION=ON`), each passing its GNU C23 dialect flag. Policy: `-O2 -g`, assertions kept, debug symbols split by the existing versioned installer. Hardening: `_FORTIFY_SOURCE=3`, PIE, full RELRO + BIND_NOW, strong stack protector, stack-clash protection, `-fcf-protection=full`, non-executable stack, each requested explicitly and probed individually. Unsupported items are reported by configure/CMake, never silently dropped. GCC 14's `-fhardened` bundle is deliberately not used: Ubuntu's compiler spec already defines `_FORTIFY_SOURCE`, so `-Whardened` rejects the bundle and the branch was unreachable on every supported distribution.
- **Unknown configure options are fatal in `configure.ac` itself** (`enable_option_checking=fatal` default), not only when a caller remembers the flag. `--enable-option-checking=warn` still overrides deliberately. CI asserts `./configure --enable-optimizations` is rejected.
- **A real profile proof.** The profile defines `LUMINARI_PRODUCTION_PROFILE`, and `src/constants.c` embeds a marker in a `.luminari.profile` ELF section. `scripts/deployment/verify_hardened_binary.sh` requires that marker plus PIE, NX stack, RELRO/NOW, stack protector, fortify, build ID, CET notes on x86-64, and no RPATH/TEXTREL. A default `gcc -O2 -g` binary (and `/bin/true`) now fails the check on Ubuntu 24.04; `test_production_profile.sh` asserts that negative case alongside the degraded-build case.
- **Explicit LTO/PGO.** `--enable-lto`, `--with-pgo-generate=DIR`, `--with-pgo-use=PATH` and the CMake equivalents. Requesting one the compiler cannot honor is a configuration error.
- **CI.** `production-profile` matrix: Autotools with GCC, GCC 14, and Clang; CMake with GCC and Clang. Every cell builds the profile, verifies the server and test binaries, runs the full production-linked suite (Autotools `make test`, CMake full `ctest`) against that artifact, installs, and verifies `bin/luminari`. The release workflow builds and verifies the same profile. The three new scripts are executable in git.
- **CMake tests build and run on a clean tree.** The 65 CuTest sources and one shared header included `../../src/conf.h`, which only exists after an Autotools configure; they now include `conf.h` through the include path each build system already provides. The event-architecture and world-tool test scripts that preprocess game sources honor `CPPFLAGS`, and CMake points them at its generated header.
- **Benchmark.** `make -C unittests/CuTest protocol-bench` times the protocol parser under any flag set. Baseline recorded in the deployment guide: hardening cost and LTO gain are both within noise on the parser. This is parser-only evidence; the live event-core game-loop gate needs a dedicated database snapshot and is not run in CI, so LTO/PGO stay opt-in until it is measured on the hardened server.
- **Fix.** `src/bsd-snprintf.c` undefines the fortified `snprintf`/`vsnprintf` macros before its fallback definitions.
- Docs: deployment, setup, and testing guides.

## Verification

Local (Ubuntu 24.04, glibc 2.39, x86-64):

- `test_production_profile.sh` passes with GCC 13.3, Clang 18, and GCC 14.2 (the GCC 14 probe yields the identical explicit set).
- `./configure --enable-optimizations` errors out; `--enable-option-checking=warn` downgrades it to the old warning.
- Autotools + GCC `--enable-production`: 0 warnings, verifier PASS on `./luminari` and installed `bin/luminari`, `make install` alias correct, no root binary. `make test` reaches `./cutest` with 1387/1388 passing against the local dev MariaDB; the one remaining test is the syntax-check boot, which needs the isolated CI runtime that only a `*test*`/`*ci*` database name can create, and this host has no privileges to create one. The hardened binary's own `-c` boot ran through DB init and world loading against the dev `lib/`. The CI cells have that runtime.
- CMake + GCC `LUMINARI_PRODUCTION=ON -DBUILD_TESTS=ON` with `src/conf.h` moved away: configures and builds `cutest`, 0 warnings, verifier PASS on `build/bin/luminari` and `build/cutest`, all 21 `ctest` entries pass (syntax boot skipped for the same reason), installed `bin/luminari` verified.
- `make -C unittests/CuTest protocol-parser`: 31/31.
- Verifier rejects `/bin/true` and a default-flag helper.

Not verified locally: non-x86 architectures.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in production build profile for Autotools and CMake, with optional LTO and PGO support.
  * Added hardened binary verification for production builds and deployments.
  * Added protocol-parser benchmarking with configurable compiler flags and iterations.

* **Bug Fixes**
  * Configuration now rejects unknown options by default.
  * Improved compatibility for fortified C library implementations.

* **Documentation**
  * Expanded build, deployment, and testing guides with production profile and verification instructions.

* **Tests**
  * Added comprehensive production-profile validation across supported compilers and build systems.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,11 +42,14 @@ jobs:
           cp src/mud_options.example.h src/mud_options.h
           cp src/vnums.example.h src/vnums.h
 
-      - name: Build with autotools
+      - name: Build the production profile with autotools
         run: |
           autoreconf -fvi
-          ./configure
+          ./configure --enable-production
           make -j"$(nproc)"
+
+      - name: Verify the hardened release binary
+        run: scripts/deployment/verify_hardened_binary.sh ./luminari
 
       - name: Generate changelog
         id: changelog
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,7 @@ on:
       - 'unittests/**'
       - 'scripts/world/**'
       - 'scripts/ci/**'
+      - 'scripts/deployment/**'
       - 'scripts/operations/**'
       - 'scripts/development/generate-web-guides.sh'
       - 'scripts/events/**'
@@ -38,6 +39,7 @@ on:
       - 'unittests/**'
       - 'scripts/world/**'
       - 'scripts/ci/**'
+      - 'scripts/deployment/**'
       - 'scripts/operations/**'
       - 'scripts/development/generate-web-guides.sh'
       - 'scripts/events/**'
@@ -411,6 +413,101 @@ jobs:
 
       - name: Run the authoritative test entry point
         run: make test-all
+
+  production-profile:
+    name: Production profile (${{ matrix.build }}, ${{ matrix.cc }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        build: [autotools, cmake]
+        cc: [gcc, clang]
+        include:
+          # GCC 14 is the newest GCC on the runner image; it proves the explicit
+          # hardening set on a compiler generation CI would otherwise never see.
+          - build: autotools
+            cc: gcc-14
+    services:
+      mariadb:
+        image: mariadb:10.11
+        env:
+          MARIADB_ROOT_PASSWORD: test_root_password
+          MARIADB_DATABASE: luminari_test
+          MARIADB_USER: luminari_test
+          MARIADB_PASSWORD: test_password
+        ports:
+          - 3306:3306
+        options: >-
+          --health-cmd="healthcheck.sh --connect --innodb_initialized"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=10
+    env:
+      CC: ${{ matrix.cc }}
+      LUMINARI_TEST_MYSQL_ENABLE: '1'
+      LUMINARI_TEST_MYSQL_HOST: 127.0.0.1
+      LUMINARI_TEST_MYSQL_USER: luminari_test
+      LUMINARI_TEST_MYSQL_PASSWORD: test_password
+      LUMINARI_TEST_MYSQL_DATABASE: luminari_test
+      LUMINARI_TEST_MYSQL_PORT: '3306'
+      LUMINARI_TEST_DATA_DIR: ${{ github.workspace }}/.ci-runtime/lib
+      LUMINARI_TEST_CONFIG_FILE: .ci-runtime/lib/etc/config
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            autoconf automake build-essential clang cmake gcc-14 libcurl4-openssl-dev \
+            libevent-dev libgd-dev libjson-c-dev libmariadb-dev libssl-dev libtool \
+            mariadb-client pandoc ripgrep zlib1g-dev
+
+      - name: Copy config headers
+        run: |
+          cp src/campaign.example.h src/campaign.h
+          cp src/mud_options.example.h src/mud_options.h
+          cp src/vnums.example.h src/vnums.h
+
+      - name: Prepare isolated test runtime
+        run: scripts/ci/prepare_test_runtime.sh "$LUMINARI_TEST_DATA_DIR"
+
+      - name: Build, verify, and test the hardened artifact (autotools)
+        if: matrix.build == 'autotools'
+        run: |
+          autoreconf -fvi
+          # The retired no-op profile option must be rejected, not warned about.
+          if ./configure --enable-optimizations >configure-unknown.log 2>&1; then
+            cat configure-unknown.log
+            echo "configure accepted an unknown option" >&2
+            exit 1
+          fi
+          grep -q 'unrecognized options: --enable-optimizations' configure-unknown.log
+          # The profile must be the last word over a caller's CFLAGS/LDFLAGS.
+          ./configure --enable-production CFLAGS='-O1 -D_FORTIFY_SOURCE=2' LDFLAGS='-Wl,-z,lazy' \
+            >configure-override.log 2>&1
+          grep -Eq '^CFLAGS = -O1 -D_FORTIFY_SOURCE=2 .*-D_FORTIFY_SOURCE=3' Makefile
+          grep -Eq '^LDFLAGS = -Wl,-z,lazy .*-Wl,-z,now' Makefile
+          ./configure --enable-production
+          make -j"$(nproc)"
+          scripts/deployment/verify_hardened_binary.sh ./luminari
+          make test
+          make install
+          scripts/deployment/verify_hardened_binary.sh bin/luminari
+          test -L bin/luminari
+          test ! -e luminari
+
+      - name: Build, verify, and test the hardened artifact (cmake)
+        if: matrix.build == 'cmake'
+        run: |
+          cmake -S . -B build -DLUMINARI_PRODUCTION=ON -DBUILD_TESTS=ON
+          cmake --build build -j"$(nproc)"
+          scripts/deployment/verify_hardened_binary.sh build/bin/luminari
+          scripts/deployment/verify_hardened_binary.sh build/cutest
+          ctest --test-dir build --output-on-failure
+          cmake --install build
+          scripts/deployment/verify_hardened_binary.sh bin/luminari
 
   sanitizers:
     name: ASan, UBSan, and protocol fuzzing
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,6 +152,67 @@ foreach(HEADER
     check_include_file("${HEADER}" HAVE_${_upper_name})
 endforeach()
 
+# ========== Production build profile ==========
+# scripts/deployment/production_profile.sh is the single source of truth shared
+# with configure.ac.  It feature-detects the optimization and hardening flags
+# against the same GNU C23 dialect flag the build uses and reports anything the
+# toolchain rejects.  LTO and PGO are explicit, measured profiles and never
+# default.
+option(LUMINARI_PRODUCTION
+       "Build the optimized and hardened production profile (-O2 -g plus feature-detected hardening)"
+       OFF)
+option(LUMINARI_LTO "Add link-time optimization to the build (explicit, measured profile)" OFF)
+set(LUMINARI_PGO_GENERATE "" CACHE PATH
+    "Instrument the build to write profile data into this directory")
+set(LUMINARI_PGO_USE "" CACHE PATH
+    "Optimize using previously collected profile data at this path")
+
+set(PRODUCTION_PROFILE_ARGS)
+if (LUMINARI_PRODUCTION)
+    if (CMAKE_BUILD_TYPE AND NOT CMAKE_BUILD_TYPE STREQUAL "None")
+        message(FATAL_ERROR
+            "LUMINARI_PRODUCTION owns the optimization and debug-symbol policy; "
+            "leave CMAKE_BUILD_TYPE unset instead of '${CMAKE_BUILD_TYPE}'")
+    endif()
+    list(APPEND PRODUCTION_PROFILE_ARGS --production)
+endif()
+if (LUMINARI_LTO)
+    list(APPEND PRODUCTION_PROFILE_ARGS --lto)
+endif()
+if (LUMINARI_PGO_GENERATE)
+    list(APPEND PRODUCTION_PROFILE_ARGS --pgo-generate ${LUMINARI_PGO_GENERATE})
+endif()
+if (LUMINARI_PGO_USE)
+    list(APPEND PRODUCTION_PROFILE_ARGS --pgo-use ${LUMINARI_PGO_USE})
+endif()
+
+set(PRODUCTION_SUPPORTED "")
+if (PRODUCTION_PROFILE_ARGS)
+    execute_process(
+        COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/scripts/deployment/production_profile.sh
+                --cc ${CMAKE_C_COMPILER}
+                --cflags "${CMAKE_C23_EXTENSION_COMPILE_OPTION}"
+                ${PRODUCTION_PROFILE_ARGS}
+        OUTPUT_VARIABLE PRODUCTION_PROFILE_OUTPUT
+        ERROR_VARIABLE PRODUCTION_PROFILE_ERROR
+        RESULT_VARIABLE PRODUCTION_PROFILE_RESULT)
+    if (NOT PRODUCTION_PROFILE_RESULT EQUAL 0)
+        message(FATAL_ERROR "Production build profile failed: ${PRODUCTION_PROFILE_ERROR}")
+    endif()
+    foreach (key CFLAGS LDFLAGS SUPPORTED UNSUPPORTED)
+        string(REGEX MATCH "PRODUCTION_${key}=([^\n]*)" _ "${PRODUCTION_PROFILE_OUTPUT}")
+        set(PRODUCTION_${key} "${CMAKE_MATCH_1}")
+    endforeach()
+    string(APPEND CMAKE_C_FLAGS " ${PRODUCTION_CFLAGS}")
+    string(APPEND CMAKE_EXE_LINKER_FLAGS " ${PRODUCTION_LDFLAGS}")
+    message(STATUS "Production profile flags: ${PRODUCTION_CFLAGS} ${PRODUCTION_LDFLAGS}")
+    if (PRODUCTION_UNSUPPORTED)
+        message(WARNING
+            "Hardening features not supported by ${CMAKE_C_COMPILER} and left out: "
+            "${PRODUCTION_UNSUPPORTED}")
+    endif()
+endif()
+
 if (HAVE_STDLIB_H AND HAVE_STDIO_H AND HAVE_STRING_H)
     set(STDC_HEADERS 1)
 endif()
@@ -1101,6 +1162,12 @@ if (BUILD_TESTS)
         native-event-architecture
         event-pubsub-retirement
         PROPERTIES WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+    # These scripts preprocess game sources; point them at the generated conf.h.
+    set_tests_properties(
+        demand-driven-event-architecture
+        native-event-architecture
+        PROPERTIES ENVIRONMENT
+            "CC=${CMAKE_C_COMPILER};CPPFLAGS=-I${CMAKE_CURRENT_BINARY_DIR}")
 
     add_test(NAME world-tools
         COMMAND Python3::Interpreter -m unittest discover
@@ -1120,10 +1187,13 @@ if (BUILD_TESTS)
         world-tool-wrapper
         PROPERTIES
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+    # The world tests preprocess game sources; point them at the generated conf.h.
+    set(WORLD_TOOLS_TEST_ENV "CPPFLAGS=-I${CMAKE_CURRENT_BINARY_DIR}")
     if (TARGET rol_mob_calculator)
-        set_tests_properties(world-tools PROPERTIES
-            ENVIRONMENT "ROL_MOB_CALCULATOR=$<TARGET_FILE:rol_mob_calculator>")
+        list(APPEND WORLD_TOOLS_TEST_ENV
+            "ROL_MOB_CALCULATOR=$<TARGET_FILE:rol_mob_calculator>")
     endif()
+    set_tests_properties(world-tools PROPERTIES ENVIRONMENT "${WORLD_TOOLS_TEST_ENV}")
 
     add_test(NAME autorun-supervision
         COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/autorun/test_autorun_supervision.sh)
@@ -1138,6 +1208,11 @@ if (BUILD_TESTS)
         COMMAND bash
             ${CMAKE_CURRENT_SOURCE_DIR}/scripts/deployment/test_binary_name_static.sh)
     set_tests_properties(binary-name-static PROPERTIES
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+    add_test(NAME production-profile
+        COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/scripts/deployment/test_production_profile.sh)
+    set_tests_properties(production-profile PROPERTIES
+        ENVIRONMENT "CC=${CMAKE_C_COMPILER}"
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
 
     add_test(NAME healthcheck
@@ -1242,6 +1317,8 @@ message(STATUS "  C Standard: GNU C23")
 message(STATUS "  C Compiler: ${CMAKE_C_COMPILER_ID} ${CMAKE_C_COMPILER_VERSION}")
 message(STATUS "  MySQL Client: ${LUMINARI_MYSQL_TARGET}")
 message(STATUS "  Build Type: ${CMAKE_BUILD_TYPE}")
+message(STATUS "  Production Profile: ${LUMINARI_PRODUCTION} (${PRODUCTION_SUPPORTED})")
+message(STATUS "  LTO: ${LUMINARI_LTO}")
 message(STATUS "  Build Utilities: ${BUILD_UTILS}")
 message(STATUS "  Build Tests: ${BUILD_TESTS}")
 message(STATUS "  Developer Mode: ${DEVELOPER_MODE}")
--- a/Makefile.am
+++ b/Makefile.am
@@ -6,6 +6,8 @@ SUBDIRS = util
 
 # Additional compiler flags
 # Revert to project defaults: do NOT force global feature defines here.
+# The production profile lives in CFLAGS/LDFLAGS (configure appends it after
+# the user's flags) rather than here, where user CFLAGS would override it.
 AM_CFLAGS = @CSTD_CFLAGS@ @MYFLAGS@ @LIBEVENT_CFLAGS@
 
 
@@ -824,6 +826,7 @@ EXTRA_DIST = \
 	unittests/README_unittests.md \
 	unittests/CuTest/Makefile \
 	unittests/CuTest/README.txt \
+	unittests/CuTest/bench_protocol_parser.c \
 	unittests/CuTest/fuzz_protocol_parser.c \
 	unittests/CuTest/fuzz_corpus/plain_command.txt \
 	unittests/CuTest/fuzz_corpus/gmcp_like.txt \
@@ -857,8 +860,11 @@ EXTRA_DIST = \
 	scripts/deployment/generate_build_identity.sh \
 	scripts/deployment/install_versioned_binary.sh \
 	scripts/deployment/move_bin.sh \
+	scripts/deployment/production_profile.sh \
 	scripts/deployment/setup.sh \
+	scripts/deployment/test_production_profile.sh \
 	scripts/deployment/test_versioned_binary_install.sh \
+	scripts/deployment/verify_hardened_binary.sh \
 	scripts/development/check_local_port_allocations.sh \
 	scripts/development/dev_create_test_character.sh \
 	scripts/development/dev_kohdee_login_smoke.sh \
@@ -1024,7 +1030,7 @@ distcheck-archive:
 
 # Run unit tests
 test: check check-build-parity test-help-sync test-autorun-supervision test-versioned-binary-install \
-	test-binary-name-static test-healthcheck test-sql-interpolation \
+	test-binary-name-static test-healthcheck test-sql-interpolation test-production-profile \
 	test-background-help-entries test-demand-driven-architecture test-event-admission \
 	test-event-core-performance-gate test-native-event-architecture test-pubsub-retirement \
 	test-vessel-tooling
@@ -1036,7 +1042,7 @@ test: check check-build-parity test-help-sync test-autorun-supervision test-vers
 	test-help-sync-integration test-protocol test-protocol-fuzz test-character-rename-static \
 	test-character-rename-schema test-background-help-entries test-autorun-supervision \
 	test-versioned-binary-install test-binary-name-static test-healthcheck \
-	test-sql-interpolation \
+	test-sql-interpolation test-production-profile \
 	test-demand-driven-architecture test-event-admission test-native-event-architecture \
 	test-event-core-performance-gate test-pubsub-retirement \
 	test-vessel-tooling
@@ -1100,6 +1106,9 @@ test-versioned-binary-install:
 
 test-binary-name-static:
 	$(SHELL) ./scripts/deployment/test_binary_name_static.sh
+
+test-production-profile:
+	CC="$(CC)" ./scripts/deployment/test_production_profile.sh
 
 test-healthcheck:
 	./scripts/operations/test_healthcheck.sh
--- a/configure.ac
+++ b/configure.ac
@@ -9,6 +9,12 @@ AC_SUBST(CRYPTLIB)
 AC_SUBST(CSTD_CFLAGS)
 
 AC_CONFIG_HEADERS([src/conf.h:conf.h.in])
+
+dnl Unknown configure options are fatal by default.  The production profile was
+dnl once requested through an option configure.ac never defined, and the build
+dnl silently used the default flags; a caller may still pass
+dnl --enable-option-checking=no or =warn to override this deliberately.
+m4_divert_text([DEFAULTS], [enable_option_checking=fatal])
 AC_DEFINE([CIRCLE_UNIX], [1], [Define for Unix-based systems])
 
 dnl Find the 'more' program
@@ -94,6 +100,69 @@ else
   dnl We aren't using gcc so we can't assume any special flags.
   MYFLAGS=""
 
+fi
+
+dnl Production build profile.  scripts/deployment/production_profile.sh is the
+dnl single source of truth shared with CMake: it feature-detects the
+dnl optimization and hardening flags and reports anything the toolchain
+dnl rejects.  LTO and PGO are explicit, measured profiles and never default.
+AC_ARG_ENABLE([production],
+  [AS_HELP_STRING([--enable-production],
+    [build the optimized and hardened production profile (-O2 -g plus feature-detected hardening)])],
+  [], [enable_production=no])
+AC_ARG_ENABLE([lto],
+  [AS_HELP_STRING([--enable-lto],
+    [add link-time optimization to the build (explicit, measured profile)])],
+  [], [enable_lto=no])
+AC_ARG_WITH([pgo-generate],
+  [AS_HELP_STRING([--with-pgo-generate=DIR],
+    [instrument the build to write profile data into DIR])],
+  [], [with_pgo_generate=no])
+AC_ARG_WITH([pgo-use],
+  [AS_HELP_STRING([--with-pgo-use=PATH],
+    [optimize using previously collected profile data at PATH])],
+  [], [with_pgo_use=no])
+
+PRODUCTION_CFLAGS=""
+PRODUCTION_LDFLAGS=""
+production_profile_args=""
+AS_IF([test "x$enable_production" != xno],
+  [production_profile_args="$production_profile_args --production"])
+AS_IF([test "x$enable_lto" != xno],
+  [production_profile_args="$production_profile_args --lto"])
+AS_IF([test "x$with_pgo_generate" != xno],
+  [production_profile_args="$production_profile_args --pgo-generate $with_pgo_generate"])
+AS_IF([test "x$with_pgo_use" != xno],
+  [production_profile_args="$production_profile_args --pgo-use $with_pgo_use"])
+
+if test "x$production_profile_args" != x; then
+  AC_MSG_CHECKING([the production build profile for $CC])
+  production_profile_output=`$SHELL "$srcdir/scripts/deployment/production_profile.sh" \
+    --cc "$CC" --cflags "$CSTD_CFLAGS" $production_profile_args 2>conftest.profile.err`
+  production_profile_status=$?
+  if test $production_profile_status -ne 0; then
+    AC_MSG_RESULT([failed])
+    cat conftest.profile.err >&AS_MESSAGE_LOG_FD
+    production_profile_error=`cat conftest.profile.err`
+    rm -f conftest.profile.err
+    AC_MSG_ERROR([$production_profile_error])
+  fi
+  rm -f conftest.profile.err
+  PRODUCTION_CFLAGS=`printf '%s\n' "$production_profile_output" | sed -n 's/^PRODUCTION_CFLAGS=//p'`
+  PRODUCTION_LDFLAGS=`printf '%s\n' "$production_profile_output" | sed -n 's/^PRODUCTION_LDFLAGS=//p'`
+  production_supported=`printf '%s\n' "$production_profile_output" | sed -n 's/^PRODUCTION_SUPPORTED=//p'`
+  production_unsupported=`printf '%s\n' "$production_profile_output" | sed -n 's/^PRODUCTION_UNSUPPORTED=//p'`
+  AC_MSG_RESULT([$PRODUCTION_CFLAGS $PRODUCTION_LDFLAGS])
+  AC_MSG_NOTICE([production hardening enabled: ${production_supported:-none}])
+  for production_feature in $production_unsupported; do
+    AC_MSG_WARN([hardening feature not supported by $CC and left out: $production_feature])
+  done
+  dnl Automake compiles with $(AM_CFLAGS) $(CFLAGS), so anything carried in
+  dnl AM_CFLAGS loses to a user CFLAGS such as -D_FORTIFY_SOURCE=2.  Append the
+  dnl profile to CFLAGS/LDFLAGS themselves, after the user's values, so the
+  dnl repository policy is the last word exactly as it is under CMake.
+  CFLAGS="$CFLAGS $PRODUCTION_CFLAGS"
+  LDFLAGS="$LDFLAGS $PRODUCTION_LDFLAGS"
 fi
 
 dnl Checks for libraries.  We check for the library only if the function is
--- a/scripts/deployment/deploy.sh
+++ b/scripts/deployment/deploy.sh
@@ -10,7 +10,7 @@
 #   -h, --help        Show help message
 #   --auto            Automated setup with defaults (no prompts)
 #   -d, --dev         Development mode (includes debug tools)
-#   -p, --prod        Production mode (optimized build)
+#   -p, --prod        Production profile (optimized, hardened, verified build)
 #   --skip-deps       Skip dependency installation
 #   --skip-db         Skip database setup (NOT RECOMMENDED - database is required)
 #   --install-systemd Install/update only the canonical systemd unit
@@ -292,6 +292,16 @@ EOF
     print_msg "$GREEN" "Database setup complete!"
 }
 
+# Fail the deployment when the installed executable lacks a required
+# production hardening property.
+verify_production_binary() {
+    print_msg "$GREEN" "Verifying production hardening of bin/luminari..."
+    if ! "$PROJECT_ROOT/scripts/deployment/verify_hardened_binary.sh" "$PROJECT_ROOT/bin/luminari"; then
+        print_msg "$RED" "Production binary failed the hardening check"
+        exit 1
+    fi
+}
+
 # Function to build the project
 build_project() {
     print_header "Building LuminariMUD"
@@ -317,10 +327,12 @@ build_project() {
             chmod +x unittests/CuTest/make-tests.sh
         fi
 
-        # Configure
+        # Configure.  configure.ac makes unknown options fatal, so a
+        # misspelled or removed profile can never silently fall back to the
+        # default flags.
         print_msg "$GREEN" "Running configure..."
         if [[ "$BUILD_TYPE" == "production" ]]; then
-            ./configure --enable-optimizations
+            ./configure --enable-production
         else
             ./configure
         fi
@@ -342,6 +354,10 @@ build_project() {
 
         print_msg "$GREEN" "Build and install complete: bin/luminari"
 
+        if [[ "$BUILD_TYPE" == "production" ]]; then
+            verify_production_binary
+        fi
+
     elif [[ -f CMakeLists.txt ]]; then
         print_msg "$GREEN" "Building with CMake..."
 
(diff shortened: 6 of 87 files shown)


moshehbenavraham (commented):
**Review event:** COMMENT (GitHub forbids REQUEST_CHANGES on the PR author's own pull request). Treat this as a request-changes review: do not merge until the blocking items are fixed.

# Adversarial review of PR 153 vs issue #93

Head `06225c55dead88f9b0b1f3c150d2d836180b1aba` vs base `0802522e997e90a36b89918c0b390aae19f5b6de` (17 files). This review tries to refute the PR's claims against issue #93: a repository-owned production profile, no silent unknown-option fallback, Autotools/CMake equivalence, reported-not-dropped unsupported hardening, ELF verification that fails on regression, release tests on the hardened artifact, retained crash symbols, and LTO/PGO remaining explicit/measured.

**Verdict: request changes.** The profile probe and Autotools flag wiring are real, but the contract is not enforced on a clean checkout. Every new production-profile CI cell is red, and the change also breaks the existing `make test` path. Unique marker for this review: `ADVERSARIAL CHECK: unknown-option fatal behavior is caller-owned, not configure.ac-owned.`

## Summary

`production_profile.sh` is a coherent shared probe: GCC 13 selects the per-flag set (`-O2 -g`, `_FORTIFY_SOURCE=3`, PIE, RELRO/NOW, strong stack protector, stack-clash, `cf-protection=full`, noexecstack) and prints `PRODUCTION_UNSUPPORTED` rather than dropping flags quietly. Autotools `AM_CFLAGS`/`AM_LDFLAGS` do put those flags on `luminari` (CI link line contains `-pie -Wl,-z,relro -Wl,-z,now -Wl,-z,noexecstack`). CMake appends the same strings to `CMAKE_C_FLAGS` / `CMAKE_EXE_LINKER_FLAGS`. LTO/PGO stay opt-in and fail configure when requested and unsupported.

That is not enough. The three new scripts are git mode `100644`. Direct execution is `EACCES` / exit 126. CI never runs the ELF verifier. `make test` now depends on `./scripts/deployment/test_production_profile.sh` and fails on a GitHub checkout. The CMake job cannot compile `cutest` on a clean tree because tests include `src/conf.h` while CMake writes `build/conf.h`. The verifier also PASSes `/bin/true` and a plain `gcc -O2 -g` helper on Ubuntu 24.04, so it does not prove a repository-owned flag contract over distro defaults.

## Blocking defects

### 1. New scripts are not executable in git — CI and `make test` die with exit 126

- **Claim hit:** "A CI job inspects the linked binary and fails when required properties regress." "Release tests run against the exact hardened artifact." "`deploy.sh --prod` selects a real profile."
- **Evidence:** `git ls-files -s` records `production_profile.sh`, `test_production_profile.sh`, and `verify_hardened_binary.sh` as `100644`. Existing siblings such as `deploy.sh` are `100755`. GitHub checkout restores 0644. Direct exec of a 0644 copy returns 126 (`Permission denied`).
- **Live CI:**
  - Production profile (autotools, gcc/clang): `scripts/deployment/verify_hardened_binary.sh: Permission denied` after a successful link. The verifier never ran. The production-linked suite never ran.
  - Behavioral tests: `CC="gcc" ./scripts/deployment/test_production_profile.sh` → `Permission denied` → `make: *** [test-production-profile] Error 126`. This PR therefore regresses the pre-existing `make test` job, not only the new matrix.
  - Release workflow step `scripts/deployment/verify_hardened_binary.sh ./luminari` has the same invocation.
- **Why it slips locally:** the working tree can be `0755` while the index stays `100644` (`core.filemode=false` here). `git status` stays clean.
- **Call sites that exec the blob:** `.github/workflows/test.yml` autotools/cmake steps, `.github/workflows/release.yml`, `deploy.sh` `verify_production_binary`, `Makefile.am` `test-production-profile`, CMake `add_test(NAME production-profile COMMAND .../test_production_profile.sh)`.
- **Contrast:** `configure.ac` and CMake invoke `production_profile.sh` via `$SHELL` / `bash`, so configure succeeds without `+x`. `test-binary-name-static` already uses `$(SHELL)` because that helper is also `100644`. The new target does not.
- **Fix:** `git add --chmod=+x` on the three scripts (and keep a `bash`/`$(SHELL)` prefix if you want 0644 helpers). Do not treat a local `0755` working tree as proof.

### 2. CMake production-profile CI cannot build `cutest` on a clean tree

- **Claim hit:** Autotools/CMake equivalence; "runs the production-linked suite against that exact artifact."
- **Hunk:** `.github/workflows/test.yml` `cmake -S . -B build -DLUMINARI_PRODUCTION=ON -DBUILD_TESTS=ON` then `cmake --build`.
- **CI:** `unittests/CuTest/test_syntax_check_boot.c:3: #include "../../src/conf.h"` → `No such file or directory`. `gmake: *** [all] Error 2` before any verify/ctest.
- **Why:** CMake `configure_file(... ${CMAKE_CURRENT_BINARY_DIR}/conf.h)` and `include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})`. Relative `../../src/conf.h` does not use `-I`. Autotools writes `src/conf.h`. This host has a leftover `src/conf.h` from Autotools, which is how a local `BUILD_TESTS=ON` run can pass and CI cannot. This job is the first CMake `BUILD_TESTS=ON` path in `test.yml`.
- **Follow-on:** even after `conf.h`, the same 100644 `verify_hardened_binary.sh` invocations would 126, and `add_test(production-profile)` execs the 644 helper.

### 3. CMake CI does not run release tests against the hardened server

- **Claim hit:** "Release tests run against the exact hardened artifact."
- **Hunk:** `ctest --test-dir build --output-on-failure -R 'production-cutest|production-profile'`.
- **What that regex runs:** only the two CTest names `production-cutest` (`cutest`) and `production-profile` (the helper script). It excludes `bsd-snprintf-fallback`, which the PR description lists as verified. It never exercises `luminari` as the test artifact; it would test `cutest` if `cutest` compiled.
- **Autotools side of the same job** would have run full `make test` (help-sync, event-core unit tests, vessel tooling, then `./cutest`) after verify. That is not equivalent, and it never happened because of defect 1.
- **`deploy.sh --prod`** prefers Autotools whenever `configure.ac` exists, so the CMake install-then-verify path is dead for the documented deploy entry point. CI was the CMake proof and is red.

### 4. ELF verifier false-PASSes Ubuntu default binaries

- **Claim hit:** "ELF verification that fails on regression" / repository-owned contract rather than toolchain defaults (the original #93 bug).
- **Hunk:** `scripts/deployment/verify_hardened_binary.sh` dyn-sym / note heuristics (`__stack_chk_fail`, `__*_chk`, `IBT`+`SHSTK`, `ET_DYN`+`INTERP`, `GNU_RELRO`, `BIND_NOW`/`FLAGS_1 NOW`).
- **Evidence on this host (Ubuntu 24.04, GCC 13.3, glibc 2.39):**
  - `bash verify_hardened_binary.sh /bin/true` → `result: PASS` (pie noexecstack relro bind-now stack-protector fortify-source build-id cf-protection no-rpath no-textrel).
  - `gcc -O2 -g` helper with no production flags → same PASS.
  - `test_production_profile.sh` PASSes because the positive case builds with profile flags and the negative case is a deliberately wrecked link (`-fno-PIE -no-pie -fno-stack-protector -U_FORTIFY_SOURCE -z norelro -z lazy -z execstack -rpath`). That does not prove a default Autotools binary would fail.
- **So:** the check is a floor against a maliciously stripped ELF, not a proof that `--enable-production` / `LUMINARI_PRODUCTION` was applied. On the distro CI actually uses, a non-production `-g -O2` binary is expected to pass. CET is skipped off x86-64; the negative test does not assert `cf-protection` or `build-id`.
- **Fix direction:** assert something the default spec does not already provide, or compare `readelf` output to the probe's `PRODUCTION_SUPPORTED` set, or fail CI if `PRODUCTION_CFLAGS` is missing from the actual compile/link lines. Keep the degraded case.

## Required adversarial checks

These are the issue #93 / plan checks. Each is a finding or an explicit residual that was examined.

### Unknown-option fatal behavior on non-prod configure

`ADVERSARIAL CHECK: unknown-option fatal behavior is caller-owned, not configure.ac-owned.`

`configure.ac` adds `--enable-production` but does not set `enable_option_checking=fatal`. Autoconf's default is still warn-and-continue. `./configure --enable-optimizations` (the original no-op) still emits `unrecognized options` and proceeds with default flags unless the caller passes `--enable-option-checking=fatal`.

`deploy.sh` (both prod and non-prod) and the new CI/release configure lines do pass that flag. That satisfies the narrow "`deploy.sh --prod` no longer emits an unrecognized-option warning" criterion **once defect 1 is fixed**. It does not make unknown options fatal for a documented `./configure --enable-production` that omits the extra flag. Docs (`DEPLOYMENT_GUIDE.md`) state "Unknown configure options are fatal" as if `configure.ac` owned it.

### CMake vs Autotools flag application, including production LDFLAGS on the installed server

- Autotools: `AM_CFLAGS`/`AM_LDFLAGS` + `util/Makefile.am` the same. CI autotools gcc/clang link of `luminari` includes `-pie -Wl,-z,relro -Wl,-z,now -Wl,-z,noexecstack`. LDFLAGS would have reached the binary that `make install` copies. Verifier never ran on it.
- CMake: `string(APPEND CMAKE_C_FLAGS ...)` and `CMAKE_EXE_LINKER_FLAGS`. Probe is **not** passed `--cflags` with the C23 flag (Autotools passes `--cflags "$CSTD_CFLAGS"`). Unlikely to change hardening on GCC 13; it is still an equivalence gap.
- CMake never linked `luminari` in CI (cutest compile failed). `deploy.sh --prod` never takes the CMake branch while `configure.ac` exists.
- `CMAKE_BUILD_TYPE` conflict with `LUMINARI_PRODUCTION` is a real fail-fast; good.

### `-fhardened` vs per-flag path and GCC 14 absence in CI

`ubuntu-latest` / this host: GCC 13.3, `gcc: error: unrecognized command-line option '-fhardened'`. CI configure output is the per-flag path with empty `PRODUCTION_UNSUPPORTED`. The `-fhardened` branch (cflags+ldflags, skip individual probes except noexecstack, rely on `-Whardened` + `-Werror` in the probe) is untested. PR body already admits this. Still a hole in the stated GCC 14+ policy.

### Verifier false pass/fail (dyn-sym heuristics, CET notes, non-x86)

Covered in blocking defect 4. Additional notes:
- `__stack_chk_fail` / `__*_chk` in `.dynsym` are present on ordinary Ubuntu PIE ELFs; LTO could in theory drop some `_chk` refs (false fail) but was not measured.
- CET requires both `IBT` and `SHSTK` in notes, and only if `Machine` matches `*X86-64*`. i386 is skipped. ARM leaves `cf-protection` to `PRODUCTION_UNSUPPORTED` and the verifier does not check it.
- `GNU_STACK` grep `RWE|R E` matches this `readelf -lW` `RW` line; hex flag fields were not seen here.

### CI `ctest -R` vs "release tests run against the exact hardened artifact"

Covered in blocking defect 3. Autotools `make test` is a much larger target than CMake's two-test regex. Neither job completed. `bsd-snprintf-fallback` is built with `-DBROKEN_SNPRINTF -DTEST_SNPRINTF` under the same `CMAKE_C_FLAGS` / `AM_CFLAGS`, which is the right place to catch fortify-vs-fallback breakage, but CMake CI does not run it.

### `deploy.sh` CMake/Autotools install-then-verify

Autotools: configure fatal + `--enable-production`, `make install`, then `verify_production_binary` on `bin/luminari`. Correct shape; blocked by 100644. CMake branch: `cmake -DLUMINARI_PRODUCTION=ON` (no `BUILD_TESTS`), `cmake --install`, same verify. Unreachable in this repo because Autotools is preferred when `configure.ac` exists. No CMake `BUILD_TESTS` in deploy, which is fine for a production install.

### PGO/LTO error paths

- `--pgo-generate` and `--pgo-use` mutually exclusive: yes.
- Missing `--cc` / unknown option / missing PGO path: `test_production_profile.sh` covers those for the probe.
- Requested LTO/PGO that cannot compile: `fail()` (strict=0 for LTO/PGO, so unused-argument warnings do not fail; a compiler that ignores `-flto=thin` could still "succeed"). Residual.
- PGO generate does not `mkdir` the directory. Residual.
- Clang vs GCC flag spellings (`-flto=thin` vs `-flto=auto`; Clang use is `-fprofile-use` without gcc's `-fprofile-correction`) look right and are untested here (no clang in this environment; CI clang configure succeeded with the per-flag hardening set).

### Fortify macros vs `bsd-snprintf`

`src/bsd-snprintf.c` `#undef vsnprintf` / `#undef snprintf` immediately before the fallback definitions is the right fix for Clang fortify wrappers when `BROKEN_SNPRINTF`/`TEST_SNPRINTF` rebuild the functions. The `#undef` sits outside the `dopr` `#endif` and inside the later `#ifndef HAVE_*` guards, so a normal game translation unit with `HAVE_SNPRINTF` only undefines the macros for the rest of that one file (no further `snprintf` uses). Not a blocking defect. CMake CI never ran `bsd-snprintf-fallback`.

### Issue #93 game-loop benchmark vs protocol-only bench

Issue #93 asked for "a small performance/latency benchmark suite for the game loop and hot parsers" and "Performance impact is recorded against a stable benchmark before making LTO/PGO default."

This PR adds `unittests/CuTest/bench_protocol_parser.c` and `make -C unittests/CuTest protocol-bench`. Default `PROTOCOL_BENCH_CFLAGS` is `-O2 -g`, not the production profile; docs tell you to pass profile flags. The recorded table is parser-only. The live game-loop path is the pre-existing event-core performance gate; `make test` runs `test_event_core_performance_gate.py` (unit tests of the gate), not `run_event_core_performance_gate.py` on the hardened `bin/luminari`. LTO stays opt-in on parser noise. That meets "do not default LTO/PGO" and does not meet "game loop + hot parsers evidence." Residual / incomplete vs #93, not a runtime bug.

### Docs

`DEPLOYMENT_GUIDE.md`, `SETUP_AND_BUILD_GUIDE.md`, and `TESTING_GUIDE.md` describe the intended contract (fatal unknown options, CI matrix, verifier, crash symbols via versioned installer, protocol-bench). They overclaim relative to this head: unknown options are not fatal in `configure.ac`; CI does not verify or test the hardened artifact; `make test` includes `test-production-profile` which 126s on a clone. Crash-symbol policy (`-g`, no `NDEBUG`, existing `install_versioned_binary.sh` split) is unchanged and looks consistent; this review did not re-run a full `make install` split.

### Other residuals examined

- Branch parent is `cc0cca83` (#151). Current `master` is `0802522e` (#152 yescrypt). GitHub's three-dot file list is the 17-file commit; a two-dot vs `master` also shows password-hash files. Merge/rebase will conflict in `Makefile.am` / `CMakeLists.txt`. Rebase onto `master` before merge.
- PR description/commit still carry an AI session URL. Strip it; do not add another.
- `production_profile.sh` `try_build` uses `-Werror -Wl,--fatal-warnings` for hardening probes: good (unused flags count as unsupported).
- Assertions: no `NDEBUG` in the probe. Checked.
- `.gitignore` adds `unittests/CuTest/protocol_parser_bench`. Fine.
- `util/Makefile.am` production flags: yes, and CI linked the utils with the same LDFLAGS.

## What would make this mergeable

1. `chmod +x` in git for the three new scripts (or invoke them with `bash`/`$(SHELL)` everywhere, including workflows and `deploy.sh`).
2. Confirm `make test` is green on a clean checkout (the behavioral job is the canary).
3. Make CMake `BUILD_TESTS=ON` compile `cutest` without a leftover `src/conf.h` (include the generated `conf.h`, or stop building `cutest` as part of `all` in that job and verify `build/bin/luminari` only).
4. Run the ELF verifier on the linked server **and** show that a non-production Ubuntu binary does **not** satisfy the same check, or document that the check is a distro-floor regression net rather than a profile proof.
5. Make CMake CI run the same production-linked suite Autotools claims to run, or stop claiming equivalence.
6. Own `enable_option_checking=fatal` in `configure.ac` if that claim stays in the docs.

## Issues (structured)

### Issue 1 -- Severity: bug
- File: .github/workflows/test.yml:304
- Description: Autotools production-profile CI execs `scripts/deployment/verify_hardened_binary.sh` after a successful hardened link. The script is git mode 100644, so the step is `Permission denied` (exit 126). Live jobs `Production profile (autotools, gcc)` and `(autotools, clang)` both died here. The ELF check and `make test` against that artifact never ran. The same 100644 blob is `test_production_profile.sh` and `production_profile.sh`.
- Suggestion: `git add --chmod=+x` the three new scripts. Keep invoking via `bash` if you want 0644. Re-run the matrix on a clean checkout.
- Status: open

### Issue 2 -- Severity: bug
- File: Makefile.am:1017
- Description: `test:` now lists `test-production-profile`, which runs `./scripts/deployment/test_production_profile.sh` (line 1095) without `$(SHELL)`. On GitHub checkout that 100644 helper 126s. Live `Behavioral tests (Luminari)` failed at `make: *** [test-production-profile] Error 126` after the rest of `make check` had `FAIL: 0`. This is a regression of the default test entry point, not an isolated new job.
- Suggestion: mark the script executable in git, or invoke it the way `test-binary-name-static` already invokes its 644 helper: `$(SHELL) ./scripts/deployment/test_production_profile.sh`.
- Status: open

### Issue 3 -- Severity: bug
- File: .github/workflows/test.yml:314
- Description: CMake production-profile CI uses `-DBUILD_TESTS=ON` and builds `cutest` as part of `all`. `test_syntax_check_boot.c` includes `../../src/conf.h`. CMake generates `build/conf.h`. Clean CI has no `src/conf.h`. Live cmake gcc/clang jobs: `fatal error: ../../src/conf.h: No such file or directory`. Local trees with a leftover Autotools `src/conf.h` hide this.
- Suggestion: Point those tests at the generated header, or do not build `cutest` in this job until that include is fixed. Verify `build/bin/luminari` even if tests are a later step.
- Status: open

### Issue 4 -- Severity: bug
- File: .github/workflows/test.yml:318
- Description: `ctest -R 'production-cutest|production-profile'` is not the production-linked suite and is not the installed server. It also omits `bsd-snprintf-fallback`. Combined with issue 3, CMake never verified or tested a hardened `luminari`. `deploy.sh` never uses this CMake path while `configure.ac` exists.
- Suggestion: After a successful CMake production build, run the same `cutest` + fallback tests Autotools `make test` runs, then verify the installed `bin/luminari`. Do not claim Autotools/CMake CI equivalence until that is true.
- Status: open

### Issue 5 -- Severity: bug
- File: scripts/deployment/verify_hardened_binary.sh:78
- Description: Fortify/stack-protector/PIE/RELRO/CET heuristics PASS on `/bin/true` and on a `gcc -O2 -g` helper with no production flags on Ubuntu 24.04. The check therefore does not fail when the repository profile is absent; it fails only when someone actively strips hardening. `test_production_profile.sh`'s degraded binary is that strip, not a default configure. Issue #93's original complaint was that observed hardening came from Ubuntu defaults.
- Suggestion: Tighten the oracle so a default `-g -O2` project binary fails, or record compile/link lines from the probe as the contract and treat the ELF script as a distro-floor net (and say so in the docs).
- Status: open

### Issue 6 -- Severity: bug
- File: scripts/deployment/deploy.sh:299
- Description: `verify_production_binary` execs `"$PROJECT_ROOT/scripts/deployment/verify_hardened_binary.sh"` the same way CI does. `deploy.sh --prod` on a clone from GitHub will fail at verify with exit 126 even after a correct hardened `make install`.
- Suggestion: Same `+x` / `bash` fix as issue 1.
- Status: open

### Issue 7 -- Severity: bug
- File: .github/workflows/release.yml:52
- Description: Release job runs `scripts/deployment/verify_hardened_binary.sh ./luminari` after `--enable-production`. Same 100644 EACCES. The release artifact would not be verified.
- Suggestion: Same as issue 1.
- Status: open

### Issue 8 -- Severity: suggestion
- File: configure.ac:103
- Description: ADVERSARIAL CHECK: unknown-option fatal behavior is caller-owned, not configure.ac-owned. `--enable-production` is defined; `enable_option_checking=fatal` is not. Bare `./configure --enable-optimizations` still warns and uses default flags. Docs say unknown options are fatal.
- Suggestion: Set `enable_option_checking=fatal` in `configure.ac` (or an `AC_ARG` that errors on `--enable-optimizations`) so the original no-op cannot silently return.
- Status: open

### Issue 9 -- Severity: suggestion
- File: scripts/deployment/production_profile.sh:182
- Description: `-fhardened` is the documented GCC 14+ path. CI and GCC 13.3 take the per-flag path (`unrecognized command-line option '-fhardened'`). No job uses GCC 14. `-Whardened` behavior, ldflags duplication, and verifier CET under `-fhardened` are untested.
- Suggestion: Add a GCC 14 (or newer) cell, or stop advertising `-fhardened` as the production path until one exists.
- Status: open

### Issue 10 -- Severity: suggestion
- File: unittests/CuTest/bench_protocol_parser.c:8
- Description: Issue #93 asked for game-loop and hot-parser evidence before LTO/PGO defaults. This file times the protocol fuzzer driver only. Default Makefile flags are `-O2 -g`, not the production profile. The event-core live gate is not run on the hardened installed server by this PR's CI.
- Suggestion: Either run the live gate on the hardened `bin/luminari` and record it, or narrow the #93 claim to "parser microbench + LTO stays opt-in."
- Status: open

### Issue 11 -- Severity: nit
- File: src/bsd-snprintf.c:709
- Description: `#undef vsnprintf` / `#undef snprintf` before the fallback definitions is the correct Clang `_FORTIFY_SOURCE` fix. Not a defect. CMake CI never ran `bsd-snprintf-fallback`; Autotools `make test` never reached `check` completion on GitHub because of issue 2.
- Suggestion: Keep the undef; make sure the fallback test actually runs in the production-profile jobs.
- Status: open

### Issue 12 -- Severity: suggestion
- File: CMakeLists.txt:154
- Description: CMake probe omits `--cflags` for the GNU C23 flag that Autotools passes. `CMAKE_EXE_LINKER_FLAGS` is the right bucket for `-pie` / `-z` flags. Residual equivalence gap only.
- Suggestion: Pass `--cc ${CMAKE_C_COMPILER} --cflags ${CMAKE_C_FLAGS}` (or the detected `-std=` flag) into the probe.
- Status: open

.github/workflows/test.yml
```diff
@@ -240,6 +242,83 @@ jobs:
...
+      - name: Prepare isolated test runtime
+        run: scripts/ci/prepare_test_runtime.sh "$LUMINARI_TEST_DATA_DIR"
+
+      - name: Build, verify, and test the hardened artifact (autotools)
+        if: matrix.build == 'autotools'
+        run: |
+          autoreconf -fvi
+          ./configure --enable-option-checking=fatal --enable-production
+          make -j"$(nproc)"
+          scripts/deployment/verify_hardened_binary.sh ./luminari
```
moshehbenavraham, line 494: **[bug]** This execs `verify_hardened_binary.sh` after a successful `--enable-production` link, but that script is git mode `100644`.

Live `Production profile (autotools, gcc/clang)` both died here: `Permission denied` (exit 126). The ELF check and `make test` against this artifact never ran.

**Suggestion:** `git add --chmod=+x` the three new scripts, or invoke them with `bash` the way `configure.ac` already invokes the probe.

Makefile.am
```diff
@@ -1009,7 +1014,7 @@ EXTRA_DIST = \
 
 # Run unit tests
 test: check test-help-sync test-autorun-supervision test-versioned-binary-install \
-	test-binary-name-static test-healthcheck \
+	test-binary-name-static test-healthcheck test-production-profile \
```
moshehbenavraham: **[bug]** Adding `test-production-profile` to the default `test:` target regresses every `make test` on a GitHub checkout.

Live `Behavioral tests (Luminari)`: `CC="gcc" ./scripts/deployment/test_production_profile.sh` → `Permission denied` → `make: *** [test-production-profile] Error 126` after `# FAIL:  0`.

The helper is `100644`. `test-binary-name-static` already uses `$(SHELL)` for the same reason (line 1092).

**Suggestion:** `git add --chmod=+x` or `$(SHELL) ./scripts/deployment/test_production_profile.sh`.

.github/workflows/test.yml
```diff
@@ -240,6 +242,83 @@ jobs:
...
+          make test
+          make install
+          scripts/deployment/verify_hardened_binary.sh bin/luminari
+          test -L bin/luminari
+          test ! -e luminari
+
+      - name: Build, verify, and test the hardened artifact (cmake)
+        if: matrix.build == 'cmake'
+        run: |
+          cmake -S . -B build -DLUMINARI_PRODUCTION=ON -DBUILD_TESTS=ON
```
moshehbenavraham, line 504: **[bug]** `-DBUILD_TESTS=ON` compiles `cutest` as part of `all`. `test_syntax_check_boot.c` does `#include "../../src/conf.h"`. CMake writes `build/conf.h`, not `src/conf.h`.

Live cmake gcc/clang jobs: `fatal error: ../../src/conf.h: No such file or directory`. A leftover Autotools `src/conf.h` hides this locally; CI is a clean tree. This is the first CMake `BUILD_TESTS=ON` job in `test.yml`.

**Suggestion:** Include the generated header (or do not build `cutest` here) and still verify `build/bin/luminari`.

.github/workflows/test.yml
```diff
@@ -240,6 +242,83 @@ jobs:
...
+          cmake --build build -j"$(nproc)"
+          scripts/deployment/verify_hardened_binary.sh build/bin/luminari
+          scripts/deployment/verify_hardened_binary.sh build/cutest
+          ctest --test-dir build --output-on-failure -R 'production-cutest|production-profile'
```
moshehbenavraham: **[bug]** `ctest -R 'production-cutest|production-profile'` is not the production-linked suite and is not the installed server. It also skips `bsd-snprintf-fallback`, which the PR description lists as verified.

Combined with the `conf.h` compile failure, CMake never verified or tested a hardened `luminari`. `deploy.sh --prod` never takes the CMake branch while `configure.ac` exists, so CI was the CMake proof.

**Suggestion:** Run the same `cutest` + fallback tests Autotools claims to run, then verify installed `bin/luminari`.

scripts/deployment/verify_hardened_binary.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# Stack protector: the strong protector always references __stack_chk_fail in
+# a program with local arrays, which every game source file has.
+stack_protector=0
+grep -q '__stack_chk_fail' <<< "$dynamic_symbols" && stack_protector=1
+check stack-protector "$stack_protector"
+
+# FORTIFY_SOURCE: fortified libc wrappers (excluding the stack protector's own
+# __stack_chk_* symbols) must be referenced.
+fortify=0
```
moshehbenavraham, line 92: **[bug]** These dyn-sym / note heuristics PASS on Ubuntu default ELFs, so they do not prove a repository-owned profile.

On Ubuntu 24.04 / GCC 13.3 / glibc 2.39: `bash verify_hardened_binary.sh /bin/true` → `result: PASS`, and a `gcc -O2 -g` helper with no production flags also PASSes (pie, NX, RELRO/NOW, stack protector, fortify, build-id, CET). `test_production_profile.sh` only fails a deliberately wrecked link (`-fno-PIE -no-pie -fno-stack-protector -U_FORTIFY_SOURCE -z norelro -z lazy -z execstack -rpath`).

That is the original #93 failure mode: observed hardening from distro defaults, not from this repo.

**Suggestion:** Make a default `-g -O2` project binary fail this check, or document it as a distro-floor net and assert compile/link lines instead.

scripts/deployment/deploy.sh
```diff
@@ -292,6 +292,16 @@ EOF
     print_msg "$GREEN" "Database setup complete!"
 }
 
+# Fail the deployment when the installed executable lacks a required
+# production hardening property.
+verify_production_binary() {
+    print_msg "$GREEN" "Verifying production hardening of bin/luminari..."
+    if ! "$PROJECT_ROOT/scripts/deployment/verify_hardened_binary.sh" "$PROJECT_ROOT/bin/luminari"; then
```
moshehbenavraham, line 299: **[bug]** `deploy.sh --prod` execs the same `100644` `verify_hardened_binary.sh`. A clone from GitHub will fail verify with exit 126 after a correct hardened `make install`.

**Suggestion:** `git add --chmod=+x`, or `bash "$PROJECT_ROOT/scripts/deployment/verify_hardened_binary.sh"`.

.github/workflows/release.yml
```diff
@@ -42,12 +42,15 @@ jobs:
...
-      - name: Build with autotools
+      - name: Build the production profile with autotools
         run: |
           autoreconf -fvi
-          ./configure
+          ./configure --enable-option-checking=fatal --enable-production
           make -j"$(nproc)"
 
+      - name: Verify the hardened release binary
+        run: scripts/deployment/verify_hardened_binary.sh ./luminari
```
moshehbenavraham, line 52: **[bug]** Same `100644` exec as the test workflow. The release job will not verify the hardened artifact.

**Suggestion:** `git add --chmod=+x` or `bash scripts/deployment/verify_hardened_binary.sh ./luminari`.

configure.ac
```diff
@@ -96,6 +96,65 @@ else
 
 fi
 
+dnl Production build profile.  scripts/deployment/production_profile.sh is the
+dnl single source of truth shared with CMake: it feature-detects the
+dnl optimization and hardening flags and reports anything the toolchain
+dnl rejects.  LTO and PGO are explicit, measured profiles and never default.
+AC_ARG_ENABLE([production],
```
moshehbenavraham, line 109: **[suggestion]** ADVERSARIAL CHECK: unknown-option fatal behavior is caller-owned, not configure.ac-owned.

`--enable-production` is defined here; `enable_option_checking=fatal` is not. Autoconf still warns-and-continues. `./configure --enable-optimizations` (the original no-op) still proceeds with default flags unless the caller also passes `--enable-option-checking=fatal`. `deploy.sh` and CI do pass it; a documented `./configure --enable-production` need not. Docs claim unknown options are fatal as if this file owned that.

**Suggestion:** Set `enable_option_checking=fatal` in `configure.ac`, or error specifically on `--enable-optimizations`.

scripts/deployment/production_profile.sh
```diff
@@ -0,0 +1,241 @@
...
+
+if [[ "$production" == 1 ]]; then
+  # Optimization and symbol policy: -O2 with full DWARF.  The versioned
+  # installer splits the debug information into a .debug companion keyed by
+  # ELF build ID, so production keeps symbolizable crashes without shipping a
+  # bloated executable.  Assertions stay enabled (no NDEBUG): a MUD that stops
+  # with a core file is preferable to one that keeps running on corrupt state.
+  cflags+=(-O2 -g)
+
+  if try_build "$trivial_source" 1 -O2 -fhardened; then
```
moshehbenavraham: **[suggestion]** GCC 14+ `-fhardened` is untested. CI and GCC 13.3 take the per-flag path (`unrecognized command-line option '-fhardened'`). No matrix cell uses GCC 14, so `-Whardened`, ldflags duplication, and verifier CET under `-fhardened` have no runtime proof.

**Suggestion:** Add a GCC 14 cell, or stop advertising `-fhardened` as the production path until one exists.

unittests/CuTest/bench_protocol_parser.c
```diff
@@ -0,0 +1,149 @@
+/*
+ * Hot-parser microbenchmark for the telnet/MSDP/GMCP protocol layer.
+ *
+ * Reuses the synthetic fuzz harness (its descriptor setup, stubs, and the
+ * chunked input/output driver) so the benchmark exercises exactly the paths
+ * the parser tests and fuzzer cover.  Build it with the flags under
+ * evaluation (see PROTOCOL_BENCH_CFLAGS in the Makefile) and compare the
+ * median nanoseconds per iteration between profiles.  The live game-loop
```
moshehbenavraham, line 8: **[suggestion]** Issue #93 asked for game-loop **and** hot-parser evidence before LTO/PGO defaults. This target times the protocol fuzzer driver only. Default `PROTOCOL_BENCH_CFLAGS` is `-O2 -g`, not the production profile. `make test` runs the event-core gate's unit tests, not `run_event_core_performance_gate.py` on the hardened installed server.

LTO staying opt-in is correct; the game-loop half of the #93 benchmark claim is not met.

**Suggestion:** Record the live gate on hardened `bin/luminari`, or narrow the issue claim to parser microbench + LTO stays opt-in.

src/bsd-snprintf.c
```diff
@@ -704,6 +704,11 @@ static void dopr_outch(char *buffer, size_t *currlen, size_t maxlen, char c)
 }
 #endif /* !defined(HAVE_SNPRINTF) || !defined(HAVE_VSNPRINTF) */
 
+/* A fortified C library wraps these names in macros; the fallback must define
+ * the plain functions. */
+#undef vsnprintf
```
moshehbenavraham, line 709: **[nit]** `#undef vsnprintf` / `#undef snprintf` immediately before the fallback definitions is the right Clang `_FORTIFY_SOURCE` fix. Not a defect. CMake CI never ran `bsd-snprintf-fallback`; Autotools `make test` never got past `test-production-profile` on GitHub.

**Suggestion:** Keep the undef; actually run the fallback test in the production-profile jobs.

CMakeLists.txt
```diff
@@ -114,6 +114,64 @@ if (CMAKE_COMPILER_IS_GNUCC)
...
+endif()
+if (LUMINARI_PGO_USE)
+    list(APPEND PRODUCTION_PROFILE_ARGS --pgo-use ${LUMINARI_PGO_USE})
+endif()
+
+set(PRODUCTION_SUPPORTED "")
+if (PRODUCTION_PROFILE_ARGS)
+    execute_process(
+        COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/scripts/deployment/production_profile.sh
+                --cc ${CMAKE_C_COMPILER} ${PRODUCTION_PROFILE_ARGS}
```
moshehbenavraham: **[suggestion]** Autotools passes `--cflags "$CSTD_CFLAGS"` into this probe; CMake does not. `CMAKE_EXE_LINKER_FLAGS` is the right bucket for `-pie` / `-z` and would apply to the installed server if this job linked. It did not: `cutest` failed first. `deploy.sh --prod` never reaches this CMake path while `configure.ac` exists.

**Suggestion:** Pass the detected `-std=` flag into the probe; do not claim Autotools/CMake deploy equivalence until `deploy.sh` or CI actually installs a CMake production binary.

moshehbenavraham:
Addressed every item in the adversarial review. Branch rebased onto current `master` (`1c592d8ad`, #155); the `Makefile.am` conflict with `test-sql-interpolation` is resolved by listing both targets.

**Issues 1, 2, 6, 7 (scripts are `100644`)** - `production_profile.sh`, `test_production_profile.sh`, and `verify_hardened_binary.sh` are now `100755` in the index (`git update-index --chmod=+x`; this host has `core.filemode=false`, which is how the working tree hid it). Every call site in `test.yml`, `release.yml`, `deploy.sh`, `Makefile.am`, and CMake execs them directly, so no `bash` prefixes were added.

**Issue 3 (CMake `cutest` cannot compile on a clean tree)** - Root cause fixed rather than worked around: the 65 CuTest sources included `../../src/conf.h`, an Autotools output path. They now include `conf.h`; Automake's `DEFAULT_INCLUDES` already carries `-I$(top_builddir)/src`, CMake already has `include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})`, and the standalone harness Makefile passes `-I../../src`. Verified locally by moving `src/conf.h` aside and configuring/building `cutest` with CMake.

**Issue 4 (CMake CI regex)** - The CMake cell now runs the full `ctest --test-dir build --output-on-failure` (production-cutest, bsd-snprintf-fallback, production-profile, help-sync, world tools, event gates, vessel tooling, and the install checks), matching what Autotools `make test` covers, then installs and verifies `bin/luminari`. Running that on a clean tree exposed that the event-architecture and world-tool scripts preprocess game sources with `-I src` and also assumed the Autotools `src/conf.h`; they now honor `CPPFLAGS`, which CMake sets to its generated header directory.

**Issue 5 (verifier passes `/bin/true`)** - Confirmed: on Ubuntu 24.04 every ELF property the verifier checked is a distro default, so no `readelf` heuristic could distinguish the profile. The profile now defines `LUMINARI_PRODUCTION_PROFILE`, `src/constants.c` embeds a marker in a `.luminari.profile` section under that define, and the verifier requires it. `/bin/true` and a default `gcc -O2 -g` helper now fail with `MISSING: production-profile`; `test_production_profile.sh` asserts that negative case on the same helper it uses for the positive case, in addition to the degraded-link case. The remaining property checks stay as the regression net they always were.

**Issue 8 (fatal option checking is caller-owned)** - `configure.ac` now sets `enable_option_checking=fatal` as the default (`m4_divert_text([DEFAULTS], ...)`), so bare `./configure --enable-optimizations` errors out; `--enable-option-checking=warn` still overrides deliberately. The redundant flag is removed from `deploy.sh`, both workflows, and the docs, and the Autotools CI cells assert the rejection before configuring for real.

**Issue 9 (`-fhardened` untested)** - Tested with Ubuntu's GCC 14.2 (`gcc-14` package): `-fhardened -Werror` fails with `'_FORTIFY_SOURCE' is not enabled by '-fhardened' because it was specified in '-D' or '-U'` because Ubuntu's compiler spec already defines it, so the branch silently fell through to the per-flag path on every distribution CI would ever use. The `-fhardened` branch is removed; the explicit set is identical on GCC 13, GCC 14, and Clang 18, and an `autotools + gcc-14` matrix cell now proves it on that compiler generation. Docs updated to match.

**Issue 10 (game-loop benchmark)** - Claim narrowed in the deployment guide and PR description: the recorded evidence is the parser microbenchmark; the live event-core gate needs a dedicated database snapshot, is not run in CI, has not yet been recorded under the production profile, and is stated as the precondition for ever defaulting LTO/PGO.

**Issue 11 (fallback test in production jobs)** - Autotools `make test` reaches `make check` (which runs `bsd_snprintf_fallback_test`) now that `test-production-profile` no longer 126s; CMake runs it through the full `ctest`.

**Issue 12 (CMake probe omits `--cflags`)** - CMake passes `--cflags "${CMAKE_C23_EXTENSION_COMPILE_OPTION}"` (`-std=gnu2x` on this toolchain), the same dialect flag Autotools passes.

Also: the session URL is removed from the commit and PR description.

Local verification (Ubuntu 24.04, no root, no sudo):
- `test_production_profile.sh` PASS with gcc 13.3, clang-18, gcc-14.2.
- `./configure --enable-optimizations` -> `configure: error: unrecognized options`.
- Autotools `--enable-production` + gcc: clean build, verifier PASS on `./luminari` and installed `bin/luminari`, symlink and no root artifact confirmed. `make test` reaches `./cutest`; 1387/1388 pass against the local dev MariaDB. The one failure is `Test_syntax_check_encounter_world_boots_and_cleans_up_once`, which needs the isolated CI runtime that `prepare_test_runtime.sh` only creates for a database named `*test*`/`*ci*`; the dev user cannot create one here. The hardened binary's own `-c` boot ran through DB init and world loading against the dev `lib/` (stopping only at the unpopulated world index), so that test is delegated to the CI cells, which have the MariaDB service.
- CMake `LUMINARI_PRODUCTION=ON -DBUILD_TESTS=ON` with `src/conf.h` moved away: `cutest` builds, verifier PASS on `build/bin/luminari` and `build/cutest`, all 21 `ctest` entries pass (syntax boot skipped as above), installed `bin/luminari` verified.
- `make -C unittests/CuTest protocol-parser`: 31/31.

moshehbenavraham:
@coderabbitai full review

Makefile.am
```diff
@@ -6,7 +6,8 @@ SUBDIRS = util
 
 # Additional compiler flags
 # Revert to project defaults: do NOT force global feature defines here.
-AM_CFLAGS = @CSTD_CFLAGS@ @MYFLAGS@ @LIBEVENT_CFLAGS@
+AM_CFLAGS = @CSTD_CFLAGS@ @MYFLAGS@ @LIBEVENT_CFLAGS@ @PRODUCTION_CFLAGS@
```
moshehbenavraham: **[bug]** Production flags live in `AM_CFLAGS`, and Automake compiles as `$(AM_CFLAGS) $(CFLAGS)`. User/`make`/env `CFLAGS` therefore win and can replace `_FORTIFY_SOURCE=3` with `=2` and add `-fno-stack-clash-protection` while `-DLUMINARI_PRODUCTION_PROFILE=1` remains. `deploy.sh --prod` runs `./configure --enable-production` without unsetting `CFLAGS`.

Repro (observed compile line has production flags, then hostile CFLAGS last):

```bash
make src/constants.o CFLAGS='-g -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fno-stack-clash-protection'
```

Preprocessor becomes `_FORTIFY_SOURCE 2` / `__USE_FORTIFY_LEVEL 2`. `verify_hardened_binary.sh` still PASSes (see sibling comment). CMake is not this bug: it appends production flags after user `CMAKE_C_FLAGS`.

**Suggestion:** Pin production flags last, e.g. `override CFLAGS += @PRODUCTION_CFLAGS@` and `override LDFLAGS += @PRODUCTION_LDFLAGS@` instead of (or in addition to, after user flags) putting them only in `AM_CFLAGS`/`AM_LDFLAGS`.

scripts/deployment/verify_hardened_binary.sh
```diff
@@ -0,0 +1,129 @@
...
+if grep -E '__[A-Za-z0-9_]+_chk(@|$)' <<< "$dynamic_symbols" | grep -vq '__stack_chk_'; then
```
moshehbenavraham: **[bug]** `fortify-source` is implemented as “any `__*_chk` dynsym”, which FORTIFY 2 also provides. Stack-clash is probed in `production_profile.sh` and never checked here. Combined with Autotools `$(AM_CFLAGS) $(CFLAGS)` order, a production-tagged binary can ship with `__USE_FORTIFY_LEVEL 2` and `-fno-stack-clash-protection` and still `result: PASS` because the marker and some `*_chk` symbol remain.

Same helper built with production flags then `-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fno-stack-clash-protection` PASSed this script. `test_production_profile.sh` does not assert level 3 vs 2 or stack-clash.

**Suggestion:** Do not treat any `__*_chk` as proof of `_FORTIFY_SOURCE=3`. Fail configure/cmake when `fortify-source` or `stack-clash-protection` land in `PRODUCTION_UNSUPPORTED`, and/or put production flags after user `CFLAGS` so they cannot be stripped.

moshehbenavraham:
Merged as 93d931cbc (squash) after a rebase onto `master` at e8cee6e7e (#154).

**What merged**

- One production profile contract, `scripts/deployment/production_profile.sh`, shared by `configure.ac` (`--enable-production`) and `CMakeLists.txt` (`LUMINARI_PRODUCTION=ON`): `-O2 -g`, assertions on, `_FORTIFY_SOURCE=3`, PIE, full RELRO + BIND_NOW, strong stack protector, stack-clash protection, `-fcf-protection=full`, NX stack; every feature probed individually and any unsupported one reported.
- Unknown configure options are fatal in `configure.ac` itself; CI asserts `--enable-optimizations` is rejected.
- `.luminari.profile` ELF marker plus `verify_hardened_binary.sh`, which now fails a default `-O2 -g` binary and `/bin/true`.
- `deploy.sh --prod`, `release.yml`, and a five-cell `Production profile` matrix (autotools gcc/clang/gcc-14, cmake gcc/clang) build, verify, and test the hardened artifact.
- Clang `_FORTIFY_SOURCE` fix in `bsd-snprintf.c`; CuTest sources include the generated `conf.h` so CMake `BUILD_TESTS=ON` works on a clean tree.

**Fixed in the final push**

- The draft-review finding that Automake's `$(AM_CFLAGS) $(CFLAGS)` order let a caller's `CFLAGS` (e.g. `-D_FORTIFY_SOURCE=2`) override the profile: configure now appends the profile to `CFLAGS`/`LDFLAGS` after the user's values, matching CMake, and the autotools CI cell asserts that ordering.
- The two CMake matrix cells failed only because `world-tool-docs` needs pandoc; the cell now installs `pandoc` and `ripgrep` like the other CMake jobs.
- Merge conflicts with #154/#156 resolved (`check-build-parity` and `test-production-profile` both in `make test`; build-parity check passes).

**Follow-up**

- #157: the `Clean archive, both build systems (blocking)` job from #154 fails on `master` itself (encounter-world boot and pet-persistence tests run without the isolated runtime). Pre-existing, not introduced here.
- LTO/PGO remain opt-in; the live game-loop gate on the hardened server is still unmeasured, as noted in the PR description.

PR #173, opened 2026-09-12
Mark shop rooms on ASCII maps

### LINKED ISSUE #170: Map: mark shop rooms on the ASCII map
## Proposal

Display shop rooms as a `$` symbol on the ASCII minimap (`map` and `map world`) so players can find vendors at a glance.

## Current state (verified 2026-09-12)

- `src/asciimap.c` renders rooms from sector type through `map_info[]` and `world_map_info[]`, with reserved special values `SECT_EMPTY`, `SECT_STRANGE`, and `SECT_HERE` defined after `NUM_ROOM_SECTORS` (`src/asciimap.c:45` to `47`). `MapArea()` sets each cell to `SECT(room)` and marks the player's room with `SECT_HERE`.
- `src/obj/shop.c` keeps the global `shop_index[]` and `top_shop`; shop rooms are read through `SHOP_ROOM(shop_nr, room_index)` and end with a `NOWHERE` terminator.
- Nothing in `asciimap.c` references shops today.

## Suggested implementation

1. Add `room_has_shop(room_vnum)` in `asciimap.c` that iterates `shop_index[]` up to `top_shop` and each shop's `SHOP_ROOM` list.
2. Define `SECT_SHOP` after the existing reserved values and add entries to both display arrays (for example `\tc[\ty$\tc]\tn` on the regular map and `\ty$\tn` on the world map), inserted before the terminator markers.
3. In `MapArea()`, prefer `SECT_HERE`, then `SECT_SHOP`, then the sector.

## Considerations

- The player's own room must keep the `&` marker over the shop symbol.
- One room with several shops shows a single `$`; closed shops still show (informational, not a live status).
- Shop lookup on every mapped room is acceptable at minimap size; cache if profiling shows otherwise.

## Completion criteria

- Shops render as `$` on both map modes; the player marker still wins.
- A production-linked test covers a shop room, a non-shop room, and the player standing in a shop.
- The `map` help entry is updated in both `lib/text/help/help.hlp` and the help database.

moshehbenavraham:
Shop rooms now show `[$]` on normal ASCII maps and `$` on world maps, with a Shop legend. The player keeps `&`; closed or absent keepers do not hide shop locations, and multiple shops in one room share one marker. Automatic minimaps use the same rendering. `map world` now selects world mode with the configured default distance; `map <distance> world` remains supported.

The ASCII-map and GUI-mapper help entries are updated in `lib/text/help/help.hlp` and the local development database. The repeatable migration updates the canonical `gui-map` tag and reassigns its four aliases, including recovery from an earlier application under `map`. A MariaDB regression verifies imported entries, alias ownership, prefix lookups, repeated application, exact agreement with the help file, and preservation of unrelated content. The migration is included in the CI schema manifest and source distribution.

The clean-archive harness now copies an externally supplied test config into the archive so the server can use its required relative config path.

Local validation at `8212fae0d`:
- `make test-all` with both libevent and select: passed (1,429 production-linked tests, 31 protocol-parser tests, and supporting checks per run); `make install` passed.
- `scripts/ci/check_clean_archive.sh --keep`: passed through Autotools and CMake, including tests and installations. Strict `ci-gcc` and `ci-clang` builds each passed all 26 CTest entries.
- Hardened Autotools builds with GCC 13, Clang 18, and GCC 14, plus hardened CMake builds with GCC and Clang: builds, tests, and ELF-property checks passed. Compiler builds emitted no warnings.
- ASan/UBSan with leak detection, bounded protocol fuzzing, and Valgrind: passed; Valgrind reported zero errors.
- Coverage: 27.3% lines and 17.4% branches, above the 10.50% / 7.16% CI floors.
- All 60 applicable SQL components applied to fresh master schemas; all nine help database integration tests passed.
- World-data tools, world-file/lookup smoke checks, source-distribution hygiene, build parity, formatting, hooks, and static-analysis commands passed.
- CodeQL C/C++ security-extended and Python security/quality analyses completed, with no findings on changed lines. Gitleaks found no leaks. No dependency manifests or dependency versions changed.
- Server startup via `autorun.sh` on port 4100 passed TCP, HTTP health, clean-log, and graceful-shutdown checks. The original development MUD was restored afterward.

Local environment adaptations: the machine's unpacked libevent development package was supplied through pkg-config; hardened CMake checks used `CMAKE_SKIP_RPATH=ON`; isolated database runs used `MYSQL_TCP_PORT`. GitHub CI completion is not a merge gate for this review pass.

Fixes #170.

### DIFF
--- a/sql/components/help_map_shop_markers.sql
+++ b/sql/components/help_map_shop_markers.sql
@@ -32,7 +32,7 @@ INSERT IGNORE INTO help_keywords (help_tag, keyword) VALUES
 ('automapping', 'MAPS');
 
 INSERT INTO help_entries (tag, entry, min_level, auto_generated) VALUES
-('map', 'Use map or map world for an ASCII map of nearby rooms. Shop locations show
+('gui-map', 'Use map or map world for an ASCII map of nearby rooms. Shop locations show
 $ (even when closed), and & marks your current room, including inside a shop.
 You can set the distance with map 6 normal or map 6 world. See AUTOMAPPING
 for the ASCII map and automatic minimap options.
@@ -55,10 +55,15 @@ See also:  AUTOMAPPING
 ON DUPLICATE KEY UPDATE entry=VALUES(entry), min_level=VALUES(min_level),
                         auto_generated=VALUES(auto_generated);
 
+-- Reclaim aliases from earlier imports or the previous migration's map tag.
+DELETE FROM help_keywords
+WHERE UPPER(keyword) IN ('GUI-MAP', 'MAP', 'MAPPER', 'MAPPING')
+  AND help_tag <> 'gui-map';
+
 INSERT IGNORE INTO help_keywords (help_tag, keyword) VALUES
-('map', 'GUI-MAP'),
-('map', 'MAP'),
-('map', 'MAPPER'),
-('map', 'MAPPING');
+('gui-map', 'GUI-MAP'),
+('gui-map', 'MAP'),
+('gui-map', 'MAPPER'),
+('gui-map', 'MAPPING');
 
 COMMIT;
--- a/sql/components/help_map_shop_markers.sql
+++ b/sql/components/help_map_shop_markers.sql
@@ -1,0 +1,64 @@
+-- ASCII map shop markers and command usage. Matches lib/text/help/help.hlp.
+START TRANSACTION;
+
+INSERT INTO help_entries (tag, entry, min_level, auto_generated) VALUES
+('automapping', 'Usage: map [distance] [normal|world]
+
+The map command prints an ASCII map of your surroundings. Distance can be
+1 to 12; omitting it uses the configured default. Normal mode shows rooms
+and exits. World mode shows one symbol per room. For example, use map,
+map world, map 6 normal, or map 6 world.
+
+Shop rooms appear as [$] in normal mode and $ in world mode. Your current
+room always shows &, including when you are standing in a shop. A room with
+several shops shows one $; the marker remains when a shop is closed or its
+keeper is absent. It marks a shop location, not current vendor availability.
+Other rooms keep their terrain symbols. The same shop markers appear on
+the automatic minimap. Wilderness uses its separate terrain map.
+
+Usage: toggle automap on
+
+Automap displays a minimap with room descriptions. Use toggle automap off
+to hide it. Screen-reader mode also hides automatic maps; the map command
+remains available.
+
+See also: MAP, GUI-MAP, WILDERNESS, SCREEN-READER
+', 0, 0)
+ON DUPLICATE KEY UPDATE entry=VALUES(entry), min_level=VALUES(min_level),
+                        auto_generated=VALUES(auto_generated);
+
+INSERT IGNORE INTO help_keywords (help_tag, keyword) VALUES
+('automapping', 'AUTOMAPPING'),
+('automapping', 'MAPS');
+
+INSERT INTO help_entries (tag, entry, min_level, auto_generated) VALUES
+('map', 'Use map or map world for an ASCII map of nearby rooms. Shop locations show
+$ (even when closed), and & marks your current room, including inside a shop.
+You can set the distance with map 6 normal or map 6 world. See AUTOMAPPING
+for the ASCII map and automatic minimap options.
+
+If you are using Luminari''s GUI, a mapper will appear on the top right.
+
+To use the mapper:
+start mapping - to begin actively mapping, move slower, check your map accuracy
+stop mapping - to pause active mapping, this will help protect your map during
+     fast movement
+
+The mapper does try to sync actively with the server, BUT due to network latency,
+etc...  it is very possible to manually mess up your map while you are actively
+mapping (start mapping).  You can manually modify your map by clicking, right-
+clicking and dragging elements.  You can also add labels and do various other
+functions using the mapper.
+
+See also:  AUTOMAPPING
+', 0, 0)
+ON DUPLICATE KEY UPDATE entry=VALUES(entry), min_level=VALUES(min_level),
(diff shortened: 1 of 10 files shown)


sql/components/help_map_shop_markers.sql
```diff
@@ -0,0 +1,64 @@
...
+', 0, 0)
+ON DUPLICATE KEY UPDATE entry=VALUES(entry), min_level=VALUES(min_level),
+                        auto_generated=VALUES(auto_generated);
+
+INSERT IGNORE INTO help_keywords (help_tag, keyword) VALUES
+('automapping', 'AUTOMAPPING'),
+('automapping', 'MAPS');
+
+INSERT INTO help_entries (tag, entry, min_level, auto_generated) VALUES
+('map', 'Use map or map world for an ASCII map of nearby rooms. Shop locations show
```
moshehbenavraham: **[bug] Help migration writes tag `map`; the live entry is `gui-map`.**

`lib/text/help/help.hlp` keywords are `GUI-MAP MAP MAPPER MAPPING`. `hedit` sets `entry->tag` from the first keyword, lowercased (`src/olc/hedit.c:3182-3194`), so the database row is `gui-map`. This `INSERT` targets `'map'`.

**Exploit / repro:** Import master help.hlp, apply this file, then `help gui-map` or `help map`. `search_help` matches `LOWER(hk.keyword) LIKE 'map%'` / `'gui-map%'` (`src/help.c:398`) and `handle_database_help` prints only the first row (`src/help.c:1259-1273`). The pre-existing `gui-map` body is never updated (`INSERT IGNORE` also leaves `MAP`/`GUI-MAP` attached to it). Players still get the old GUI-mapper text with no shop-marker docs. Confirmed with shipped `catalog.parse_help_hlp`: first-keyword tag is `gui-map`; this SQL never upserts that tag.

**Fix:** Upsert `'gui-map'`. Attach the four keywords there. `DELETE FROM help_keywords WHERE UPPER(keyword) IN ('GUI-MAP','MAP','MAPPER','MAPPING') AND help_tag <> 'gui-map';` so lookups cannot stick to the stale row. Keep the `automapping` upsert.

moshehbenavraham (commented):
## Finding: help migration upserts tag `map`, not the live `gui-map` row

The GUI-MAP help entry's first keyword is `GUI-MAP`. `hedit` and help-sync derive the database tag by lowercasing that first keyword (`src/olc/hedit.c:3182-3194` -> `gui-map`). This migration inserts/updates tag `map` instead, so the row players already have is never written.

`search_help` matches `LOWER(hk.keyword) LIKE LOWER(?)` (`src/help.c:398`). `handle_database_help` then displays only the first list node (`src/help.c:1259-1273`). After this SQL runs on a DB imported from help.hlp:

- tag `gui-map` still has the pre-change body
- a new tag `map` holds the shop-marker text
- keywords `MAP` and `GUI-MAP` exist on both tags (`INSERT IGNORE` does not steal them)

`HELP GUI-MAP` (the lookup `src/act.other.c` tells players to use) and `HELP MAP` therefore can still serve the old GUI mapper blurb with no shop-marker docs. Unlike `sql/components/help_command_sweep_entries.sql`, this file never `DELETE`s those keywords from the other tag.

### Repro
1. Import current master `help.hlp` (first keyword `GUI-MAP` -> tag `gui-map`, body starts "If you are using Luminari's GUI").
2. Apply `sql/components/help_map_shop_markers.sql`.
3. `SELECT tag, LEFT(entry,80) FROM help_entries WHERE tag IN ('gui-map','map');` — `gui-map` unchanged, `map` is the new text.
4. `SELECT help_tag, keyword FROM help_keywords WHERE keyword IN ('MAP','GUI-MAP');` — both tags own both keywords.
5. In-game `help gui-map` / `help map` hits `gui-map` first whenever MySQL returns that row first (equal `LENGTH(keyword)`).

Shipped-parser run is in the review evidence; `catalog.parse_help_hlp` reports tag `gui-map`, SQL upserts `map`, and the simulated post-migration `gui-map` body still lacks shop docs.

### Fix
Upsert tag `gui-map` (not `map`). Insert keywords onto `gui-map`. Before insert, `DELETE FROM help_keywords WHERE UPPER(keyword) IN ('GUI-MAP','MAP','MAPPER','MAPPING') AND help_tag <> 'gui-map';` so lookup cannot stay pinned to the stale row. Leave `automapping` as-is; that tag already matches the first keyword.

sql/components/help_map_shop_markers.sql
moshehbenavraham: Fixed in 8212fae0d. The migration now upserts `gui-map` and reclaims `GUI-MAP`, `MAP`, `MAPPER`, and `MAPPING` from other tags before inserting them. The `automapping` upsert remains unchanged.

The new isolated MariaDB regression reproduces the stale body before the fix and passes afterward. It covers an imported canonical entry and a database that already received the faulty migration, applies the corrected SQL twice, compares both bodies with `help.hlp`, verifies all six alias owners and prefix lookup results, and preserves unrelated content. All nine help database integration tests pass.

The local development database has been corrected and verified against the PR help file. I also added the missing CI/schema and distribution manifest entries and repaired the clean-archive harness's external config path. Both previously failing CI jobs now pass locally, including all 60 applicable SQL components and clean-archive builds/tests/installations through Autotools and CMake.

moshehbenavraham:
Review follow-up is complete at `8212fae0d`: the help-tag finding is fixed, verified, and resolved. Both previously failing jobs pass locally, and the remaining applicable CI checks have been run locally as summarized in the PR description. Master is already an ancestor of this branch, so no rebase was needed.

The worktree is clean, the installed binary identifies the published commit with `GIT_DIRTY=0`, and no root-level server binary remains. The PR is ready to merge; GitHub-side job completion was not used as a gate.

PR #176, opened 2026-09-13
Cover the clean-archive runtime configuration regression

### LINKED ISSUE #157: CI: clean-archive job fails on master (encounter boot and pet persistence tests)
The `Clean archive, both build systems (blocking)` job introduced by #154 fails on `master` itself, independent of any open PR.

Observed on the `master` run for e8cee6e7e (#154) and again on the PR #153 head after rebasing onto it. The two Autotools jobs that use `scripts/ci/prepare_test_runtime.sh` pass the same suite; this job runs `make test` from an untouched `git archive HEAD` without that runtime and fails two CuTest cases:

```
1) Test_syntax_check_encounter_world_boots_and_cleans_up_once: unittests/CuTest/test_syntax_check_boot.c:526: expected <0> but was <21>
2) Test_pet_persistence_legacy_schema_migration_is_idempotent: unittests/CuTest/test_database_persistence.c:807: assert failed
Runs: 1388 Passes: 1386 Fails: 2
```

Exit 21 is `luminari_main` returning non-success in the syntax-check child, i.e. the archived `lib/` cannot boot the encounter world on its own. The pet persistence case needs the isolated test database fixture the other jobs prepare.

Likely fix: have the archive job prepare the same isolated runtime (`prepare_test_runtime.sh` plus the `LUMINARI_TEST_*` variables) before `make test` and `ctest`, or make those two tests skip cleanly when no runtime is configured.

moshehbenavraham:
The clean-archive failure in #157 came from passing an absolute runtime config path to a server that requires a safe relative path. The job already prepared MariaDB and the test runtime; #173 fixed the config handoff by copying it into the exported tree. That fix is on master and its clean-archive job is passing.

Add a regression test that runs the production archive script against a temporary Git repository and records the environment at both `make test` and `ctest`. It pins the exported config path to `lib/etc/config` and covers absolute and relative config inputs, preserved database settings, enabled boot coverage, explicit local skip behavior, and a missing config failing before any build step. Wire it into both build systems, include the scripts in source distributions, and document the runtime requirements.

Validation (`eda269205`, rebased on `5232727fb`):
- `python3 scripts/ci/test_clean_archive.py -v`: 3 tests passed, including both config path forms; they also pass with `TMPDIR` behind a symlink.
- Mutation check against copies of `check_clean_archive.sh`: removing the config relocation, exporting `./lib/etc/config`, `lib//etc/config`, or `lib/etc/config/`, and swallowing a `cp` failure each make the test fail. The unchanged script passes.
- Build parity, source hygiene, ruff, commit hooks, and the push build hook passed; the checkout binary was installed.
- All 20 jobs this PR triggers (Build & Test, Source Hygiene, Integration) ran locally from the workflow YAML, in Ubuntu 24.04 containers with MariaDB 10.11 sidecars, and passed. The first sanitizer run hit a local image gap (no `g++-14` for linking libFuzzer) and passed once the image was fixed. The review follow-up comment has per-job results. GitHub Actions on `eda269205`: all 39 checks passed.
- CodeQL 2.27.0 python code-scanning suite: no results in the changed file.

Closes #157.

## Summary by CodeRabbit

* **Tests**
  * Added automated validation for clean archive runtime handoffs.
  * Checks both supported test entry points, configuration path handling, environment propagation, and explicit no-runtime mode.
  * Missing configuration files now fail validation before build steps begin.
  * Added the archive validation to the standard test workflows.

* **Documentation**
  * Expanded testing guidance with archive runtime requirements, configuration handling, and validation coverage.


scripts/ci/test_clean_archive.py
```diff
@@ -0,0 +1,161 @@
...
+                self.env["LUMINARI_TEST_CONFIG_FILE"] = path
+                result = self.run_archive()
+                self.assertEqual(result.returncode, 0, result.stdout + result.stderr)
+                records = self.records()
+                self.assertEqual([record["tool"] for record in records], ["make", "ctest"])
+                for record in records:
+                    environment = record["environment"]
+                    archived_config = Path(environment["LUMINARI_TEST_CONFIG_FILE"])
+                    self.assertFalse(archived_config.is_absolute())
+                    self.assertNotIn("..", archived_config.parts)
```
moshehbenavraham: **[Medium] Path-safety assertion is weaker than the server's `is_safe_relative_path()`: the #157 boot failure can come back with this test green**

**Location:** `scripts/ci/test_clean_archive.py:121-123`. Server side: `src/utils.c:3956` (`is_safe_relative_path`), called at `src/comm.c:356`.

The test accepts any exported config path that is not absolute and has no `..` component. The server also rejects `.` components, empty components (`//`, trailing `/`), spaces, and any character outside `[A-Za-z0-9._-]` plus `/`. A rejected path makes `luminari_main()` return `EXIT_FAILURE` before `load_config()`. The syntax-boot child then takes `_exit(21)` (`unittests/CuTest/test_syntax_check_boot.c:501`), and the suite fails at `test_syntax_check_boot.c:526` with `expected <0> but was <21>`, the exact signature in #157.

**Repro.** Change line 58 of `scripts/ci/check_clean_archive.sh` to either of the following and run this test unmodified. Both mutants pass 3/3:

```
export LUMINARI_TEST_CONFIG_FILE=./lib/etc/config
export LUMINARI_TEST_CONFIG_FILE=lib//etc/config
```

The server rejects both. The `-d` target is deliberately nonexistent; the check fires before any chdir, DB access, or port bind:

```
$ bin/luminari -f ./lib/etc/config -c -q -d /nonexistent
SYSERR: Configuration file must be a safe path relative to the library directory.
exit=1
$ bin/luminari -f lib//etc/config -c -q -d /nonexistent
SYSERR: Configuration file must be a safe path relative to the library directory.
exit=1
```

Predicate comparison (server function compiled verbatim from `src/utils.c`):

```
path                  is_safe_relative_path()   this test
lib/etc/config        ACCEPT                    ACCEPT
./lib/etc/config      REJECT                    ACCEPT
lib//etc/config       REJECT                    ACCEPT
lib/etc/config/       REJECT                    ACCEPT
```

**Fix.** Assert the server's grammar, or pin the exact contract with `assertEqual(..., "lib/etc/config")`:

```python
# Mirror is_safe_relative_path() in src/utils.c.
archived_config = environment["LUMINARI_TEST_CONFIG_FILE"]
self.assertRegex(archived_config, r"\A[A-Za-z0-9._-]+(/[A-Za-z0-9._-]+)*\Z")
self.assertFalse({".", ".."} & set(archived_config.split("/")))
self.assertLess(len(archived_config.encode()), 256)
```

Verified: kills both mutants and the relocation-removed mutant, and still passes on the current script.

scripts/ci/test_clean_archive.py
```diff
@@ -0,0 +1,161 @@
...
+            self.assertNotIn("LUMINARI_TEST_CONFIG_FILE", environment)
+            self.assertNotIn("LUMINARI_TEST_MYSQL_ENABLE", environment)
+            self.assertEqual(environment["LUMINARI_TEST_SKIP_SYNTAX_BOOT"], "1")
+
+    def test_missing_config_fails_before_tests(self):
+        self.env["LUMINARI_TEST_CONFIG_FILE"] = str(self.root / "missing-config")
+        result = self.run_archive()
+        self.assertNotEqual(result.returncode, 0)
+        self.assertFalse(self.record.exists())
+        self.assertEqual(list(self.root.glob("luminari-archive.*")), [])
```
moshehbenavraham: **[Medium] `test_missing_config_fails_before_tests` is vacuous, and the fake models a missing config the opposite way from the server**

**Location:** `scripts/ci/test_clean_archive.py:152-157`, with the fake build tool at `:37` and `:40`.

The fake `make test`/`ctest` evaluates `Path(config).read_text()` (line 37) before it opens the record file (line 40). With a missing config it raises `FileNotFoundError` and records nothing. "Non-zero exit and no record file" therefore holds whether the script fails fast or runs `autoreconf`, `configure` and the builds and only dies inside `make test`. The test cannot tell the two apart.

The real server does the opposite of the fake. Traced:
- `load_config()` logs `No lib/etc/config file, using defaults` and returns (`src/db.c:8291-8297`).
- The compiled defaults (`src/config.c:181-187`) are mortal 3001, which exists in the prepared runtime, plus immort 1204 and frozen 1202, which do not.
- `check_start_rooms()` only warns for immort and frozen and falls back to the mortal room (`src/db.c:2586-2597`).

So the encounter boot passes on defaults, and the prepared runtime config is silently ignored.

**Repro.** Swallow the copy failure in `scripts/ci/check_clean_archive.sh` line 57:

```
cp -- "$LUMINARI_TEST_CONFIG_FILE" "$work_dir/lib/etc/config" 2>/dev/null || true
```

All 3 tests still pass, including this one. In the real job the same regression turns a missing or misnamed runtime config into a green run on compiled defaults instead of a fast failure.

**Fix.** Record every tool invocation before anything can raise, and assert that no build tool ran:

```python
# BUILD_TOOL, immediately after `args = sys.argv[1:]`
with open(os.environ['ARCHIVE_TEST_RECORD'] + '.calls', 'a') as calls:
    calls.write(json.dumps([tool, *args]) + '\\n')

# test_missing_config_fails_before_tests
self.assertFalse(Path(f"{self.record}.calls").exists())
```

Verified: kills the mutant above and passes on the current script.

scripts/ci/test_clean_archive.py
```diff
@@ -0,0 +1,161 @@
...
+    Path('bin/luminari').write_text('#!/bin/sh\\nexit 0\\n')
+    Path('bin/luminari').chmod(0o755)
+"""
+
+
+class CleanArchiveTest(unittest.TestCase):
+    def setUp(self):
+        temporary = tempfile.TemporaryDirectory(prefix="test-clean-archive-")
+        self.addCleanup(temporary.cleanup)
+        self.root = Path(temporary.name)
```
moshehbenavraham: **[Low] False failure of `make test` and the `clean-archive-runtime` CTest when the temp path contains a symlink**

**Location:** `scripts/ci/test_clean_archive.py:54`. The failing asserts are at `:129-132`.

`self.root` keeps the logical `tempfile` path and is passed to the script as `TMPDIR`. `mktemp -d` therefore returns a logical path, and the script exports it as `LUMINARI_TEST_ROOT`/`LUMINARI_TEST_SPEC_WORLD_ROOT`. The fake records `Path.cwd()`, which is `getcwd()` and always physical. With any symlink in `$TMPDIR` (a symlinked `/tmp`, or `TMPDIR` under `$HOME` where `/home -> var/home` as on Fedora Atomic), the two never match. The test fails with the production script unchanged, so both required gates break on correct code.

**Repro.**

```
d=$(mktemp -d); mkdir "$d/real"; ln -s "$d/real" "$d/link"
TMPDIR="$d/link" python3 scripts/ci/test_clean_archive.py -v
```

```
FAIL: test_external_runtime_reaches_both_test_entry_points (config_path=<absolute path>)
FAIL: test_external_runtime_reaches_both_test_entry_points (config_path='isolated runtime/lib/config')
AssertionError: '<d>/link/test-clean-archive-*/luminari-archive.*' != '<d>/real/test-clean-archive-*/luminari-archive.*'
```

**Fix.**

```python
self.root = Path(temporary.name).resolve()
```

Verified: passes under a symlinked `TMPDIR`, and the mutants from the other two comments are still killed.

moshehbenavraham (commented):
## Adversarial review: 3 findings

Method: ran mutants of `scripts/ci/check_clean_archive.sh` against the new test as submitted, compiled the real `is_safe_relative_path()` on its own to compare verdicts, and traced the config path from `LUMINARI_TEST_CONFIG_FILE` through the syntax-boot child to `load_config()`.

| # | Severity | Finding | Location |
|---|---|---|---|
| 1 | Medium | Path-safety assertion is weaker than `is_safe_relative_path()`; the #157 exit-21 boot failure can come back with this test green | `scripts/ci/test_clean_archive.py:121-123` |
| 2 | Medium | `test_missing_config_fails_before_tests` is vacuous (the fake crashes before recording); the real server silently boots on compiled defaults | `scripts/ci/test_clean_archive.py:37-40`, `:152-157` |
| 3 | Low | Logical vs physical path compare fails `make test` and CTest on correct code under a symlinked `TMPDIR` | `scripts/ci/test_clean_archive.py:54`, `:129-132` |

Mutation results against this PR's test:

```
check_clean_archive.sh variant                        test_clean_archive.py
unchanged (control)                                   pass
relocation block removed                              killed (matches PR claim)
export LUMINARI_TEST_CONFIG_FILE=./lib/etc/config     SURVIVES (server rejects)
export LUMINARI_TEST_CONFIG_FILE=lib//etc/config      SURVIVES (server rejects)
cp ... 2>/dev/null || true                            SURVIVES
unchanged, TMPDIR through a symlink                   FAILS (false failure)
```

With the combined fix below, the control passes, all four mutants are killed, and the symlinked-`TMPDIR` run passes.

<details>
<summary>Combined fix (verified)</summary>

```diff
--- a/scripts/ci/test_clean_archive.py
+++ b/scripts/ci/test_clean_archive.py
@@ -24,6 +24,8 @@
 
 tool = Path(sys.argv[0]).name
 args = sys.argv[1:]
+with open(os.environ['ARCHIVE_TEST_RECORD'] + '.calls', 'a') as calls:
+    calls.write(json.dumps([tool, *args]) + '\\n')
 if tool == 'autoreconf':
     Path('configure').write_text('#!/bin/sh\\nexit 0\\n')
     Path('configure').chmod(0o755)
@@ -51,7 +53,7 @@
     def setUp(self):
         temporary = tempfile.TemporaryDirectory(prefix="test-clean-archive-")
         self.addCleanup(temporary.cleanup)
-        self.root = Path(temporary.name)
+        self.root = Path(temporary.name).resolve()
         self.repo = self.root / "source checkout"
         self.repo.mkdir()
         script = self.repo / "scripts/ci/check_clean_archive.sh"
@@ -118,9 +120,11 @@
                 self.assertEqual([record["tool"] for record in records], ["make", "ctest"])
                 for record in records:
                     environment = record["environment"]
-                    archived_config = Path(environment["LUMINARI_TEST_CONFIG_FILE"])
-                    self.assertFalse(archived_config.is_absolute())
-                    self.assertNotIn("..", archived_config.parts)
+                    # Mirror is_safe_relative_path() in src/utils.c.
+                    archived_config = environment["LUMINARI_TEST_CONFIG_FILE"]
+                    self.assertRegex(archived_config, r"\A[A-Za-z0-9._-]+(/[A-Za-z0-9._-]+)*\Z")
+                    self.assertFalse({".", ".."} & set(archived_config.split("/")))
+                    self.assertLess(len(archived_config.encode()), 256)
                     self.assertEqual(record["config"], config_text)
                     self.assertEqual(environment["LUMINARI_TEST_DATA_DIR"], str(runtime))
                     for key, value in database.items():
@@ -154,6 +158,7 @@
         result = self.run_archive()
         self.assertNotEqual(result.returncode, 0)
         self.assertFalse(self.record.exists())
+        self.assertFalse(Path(f"{self.record}.calls").exists())
         self.assertEqual(list(self.root.glob("luminari-archive.*")), [])
 
 
```

</details>

Checked, no defects:
- The Makefile.am and CMake wiring (including `WORKING_DIRECTORY`).
- `load_config()` (`src/comm.c:370`) runs before `chdir(dir)` (`src/comm.c:510`), so `lib/etc/config` resolves from the archive root under both `make test` and CTest.
- The fixture strips inherited `LUMINARI_TEST_*`/`GIT_*`, so no CI credentials reach the recorded JSON or failure output.
- The script writes only inside its own `mktemp -d` tree.

scripts/ci/test_clean_archive.py
moshehbenavraham: Fixed in eda269205. The test now pins the exported value instead of approximating the server's grammar:

```python
self.assertEqual(environment["LUMINARI_TEST_CONFIG_FILE"], "lib/etc/config")
```

I took the exact-contract option. The script exports that literal, so any other spelling fails, including every form `is_safe_relative_path()` rejects, and there is no Python copy of the C grammar to drift. Mutants re-run against the new test (the unchanged script passes):

```
check_clean_archive.sh variant                        test_clean_archive.py
export LUMINARI_TEST_CONFIG_FILE=./lib/etc/config     killed (both config-path subtests)
export LUMINARI_TEST_CONFIG_FILE=lib//etc/config      killed (both config-path subtests)
export LUMINARI_TEST_CONFIG_FILE=lib/etc/config/      killed (both config-path subtests)
relocation block removed                              killed
```

scripts/ci/test_clean_archive.py
moshehbenavraham: Fixed in eda269205 with the approach you suggested. The fake build tool appends `[tool, *args]` to `<record>.calls` before anything that can raise. The missing-config test, renamed `test_missing_config_fails_before_build`, asserts that file never appears. That supersedes the old record-file assertion, so I dropped it.

```
check_clean_archive.sh variant    test_clean_archive.py
cp ... 2>/dev/null || true        killed (test_missing_config_fails_before_build)
relocation block removed          killed (both config-path subtests and this test)
unchanged                         pass
```

I left the server's missing-config fallback alone (`load_config()` logs and uses compiled defaults). Making it fatal would change game boot behaviour outside this PR's scope. The archive script's `cp` under `set -e` is the fast-fail guard, and this test now proves it. The testing guide now says a missing config fails before any build step.

scripts/ci/test_clean_archive.py
moshehbenavraham: Fixed in eda269205: `self.root = Path(temporary.name).resolve()`. Your repro (`TMPDIR` through a symlink) now passes 3/3, and the mutants from the other two threads are still killed with a normal `TMPDIR`.

moshehbenavraham:
## Review follow-up (eda269205)

**Branch.** Rebased onto master `5232727fb` and force-pushed over `768d50afd`. The review fixes are squashed into the single PR commit, so a squash merge keeps one coherent commit.

**Review findings** (details and verification in each thread)

| # | Finding | Resolution |
|---|---|---|
| 1 | Path assertion weaker than `is_safe_relative_path()` | Pins the exported value to `lib/etc/config` |
| 2 | Missing-config test was vacuous | Fake tools log every call before anything can raise; the test asserts no build tool ran |
| 3 | False failure with a symlinked `TMPDIR` | `self.root` is resolved |

Mutation check against copies of `check_clean_archive.sh`, same harness before and after:

```
check_clean_archive.sh variant               768d50afd test       eda269205 test
unchanged                                    pass                 pass
relocation block removed                     killed               killed
export LUMINARI_TEST_CONFIG_FILE=./lib/...   SURVIVES             killed
export LUMINARI_TEST_CONFIG_FILE=lib//...    SURVIVES             killed
export LUMINARI_TEST_CONFIG_FILE=.../config/ SURVIVES             killed
cp ... 2>/dev/null || true                   SURVIVES             killed
unchanged, TMPDIR through a symlink          FAILS (false)        pass
```

**CodeRabbit** (review of `eda269205`): no actionable comments. I did not add docstrings for its Docstring Coverage pre-merge warning. The repository's Python tests do not document test methods (6 of the 587 under `scripts/` have docstrings), and merged PRs such as #174 (79.55%) and #172 (66.67%) carried the same warning.

**Local CI.** GitHub Actions results were not treated as a gate. All 20 jobs this PR triggers (Build & Test, Source Hygiene, Integration), including every matrix entry, were expanded from the workflow YAML at `eda269205`. Each job ran in its own Ubuntu 24.04 container pinned to 4 CPUs, with a MariaDB 10.11 sidecar where the job defines one, so the database and port 4100 were private to the job.

| Workflow | Job | Result | Local wall time |
|---|---|---|---|
| Build & Test | ASan, UBSan, and protocol fuzzing | pass on rerun (note 1) | 2m 43s |
| Build & Test | Autotools/CMake manifest parity | pass | 0m 02s |
| Build & Test | Behavioral tests (Luminari) | pass | 4m 05s |
| Build & Test | CMake clang (strict, blocking) | pass | 2m 10s |
| Build & Test | CMake gcc (strict, blocking) | pass | 2m 35s |
| Build & Test | Clean archive, both build systems (blocking) | pass | 3m 31s |
| Build & Test | Coverage | pass | 1m 55s |
| Build & Test | Memory check | pass | 1m 43s |
| Build & Test | Production profile (autotools, clang) | pass | 3m 37s |
| Build & Test | Production profile (autotools, gcc) | pass | 4m 18s |
| Build & Test | Production profile (autotools, gcc-14) | pass | 4m 22s |
| Build & Test | Production profile (cmake, clang) | pass | 1m 58s |
| Build & Test | Production profile (cmake, gcc) | pass | 2m 17s |
| Build & Test | Production-linked tests (libevent) | pass | 6m 16s |
| Build & Test | Production-linked tests (select) | pass | 6m 22s |
| Build & Test | World-data tools | pass | 0m 17s |
| Source Hygiene | Artifacts, encoding, and ASCII docs | pass | 0m 01s |
| Integration | Database Migration Dry-Run | pass | 1m 53s |
| Integration | Server Startup Verification | pass | 0m 53s |
| Integration | World File Validation | pass | 0m 02s |

- Clean archive: the new test passed 3/3 inside the exported tree. All 1,434 CuTests and 27/27 CTest entries passed, including `clean-archive-runtime`, with encounter boot and pet persistence enabled.
- The new test also passed in `make test` in every other Autotools job, and `clean-archive-runtime` passed in every other CTest job.
- Server startup accepted connections on 4100 and reported healthy with a healthy database. Memory check: valgrind reported 0 errors and 0 bytes definitely lost.
- Coverage: 1,434 CuTests and 31 protocol-parser tests passed, and gcovr enforced the baseline (lines 27.3%, branches 17.4%; floors 10.50% and 7.16%), matching GitHub's Coverage job on the same commit.
- Note 1: the first sanitizer run passed its 1,434-test ASan/UBSan suite but failed to link the libFuzzer harness (`cannot find -lstdc++`). My local image had `gcc-14` without `g++-14`, so clang selected a GCC 14 directory with no libstdc++; the hosted runner image ships g++-14. After adding `g++-14` to the image, the whole job passed, including the 15-second fuzzing run.
- CodeQL 2.27.0 python code-scanning suite (43 queries): no results in `scripts/ci/`. Its 4 existing results are in `scripts/help-sync/tests/test_endpoint_integration.py`, which this PR does not change.
- Not replicated locally: CodeRabbit, an external review service (its review is summarized above).

PR #179, opened 2026-09-13
Speed up test loops and CI without losing coverage

### LINKED ISSUE #178: Make the comprehensive test suite run much faster without losing coverage
## Goal

Keep every check the comprehensive test suite makes today and make the loop much faster at all three levels: one edit on the host, `make test-all` on the host, and the full CI matrix (locally in containers and on GitHub).

Plan with measurements: [docs/ongoing-projects/test-loop-speed.md](https://github.com/LuminariMUD/Luminari-Source/blob/master/docs/ongoing-projects/test-loop-speed.md)

## Measured findings (2026-09-13)

- `./cutest` takes 33.2 s, and 30.9 s of that is one test: `Test_syntax_check_encounter_world_boots_and_cleans_up_once`. The other 1447 tests take 1.5 s combined.
- That boot spends 30 s in the mob loader: `interpret_espec()` calls `affect_total()` after every E-spec line (397,085 calls) instead of once per mob (27,092). Calling it once per mob at the end of `parse_mobile()` gives a 4.5 s full-world boot and a 6.0 s cutest run, with byte-identical prototypes (dumped `aff_abils`, `real_abils`, and `points` for every mob compared with `cmp`). The live server would boot and copyover about 26 s faster too.
- The next costs are fixed sleeps: `test_autorun_supervision.sh` (21.7 s, 7.6 s of it a hardcoded 5 s fastboot sleep in `autorun.sh`) and `test_monitor_process_memory.sh` (9.5 s, run twice by `make test-all`).
- `make -j16 test` already passes with its prerequisites running in parallel. GitHub runs `make test` and `make test-all` without `-j`, so the two slowest jobs spend over six minutes compiling on one core.
- The matrix compiles the server about 30 times per PR. `LUMINARI_IO_DRIVER` is read at run time, so the libevent and select jobs build identical binaries; `compile-check` duplicates the CMake `-Werror` jobs; the clang-tidy job cannot fail.

## Phases

1. Compute mob totals once per prototype in `src/db.c` (server fix; verified equivalent).
2. Parallel `test` and `test-all` targets, `ctest -j`, `-j` in the CI test steps, drop the duplicate process-memory run.
3. Replace fixed sleeps in the autorun and process-memory tests with polling.
4. `CUTEST_FILTER=<substring>` and an automatic list of tests slower than 1 s in cutest.
5. Build hygiene: no routine `make clean`, ccache, optional `-O0` dev configure.
6. Consolidate the CI matrix to 23 builds with the same coverage; make the clang-tidy job real or delete it.
7. Prebuilt image plus shared ccache for the local container matrix.

## Expected results

| Loop | Today | After phases 1 to 4 | After all phases |
|------|-------|---------------------|------------------|
| Edit one file, rerun one test file | 3 s build + 33 s cutest | 3 s + under 1 s | same |
| `make test` on the host | about 70 s | about 10 s | same |
| `make test-all` on the host | about 1.5 min | about 15 s | same |
| Live server boot / copyover | 31 s in the mob loader | 4.5 s | same |
| GitHub `test.yml` wall | about 17 min | about 10 min | about 5 min |
| Local container matrix | about 25 min | about 20 min | 5 to 8 min |

Host numbers were measured with the changes applied and then reverted; the matrix and GitHub numbers are estimates from per-step timings in the document.

Phases 1 to 4 fit in one small PR; phases 5 to 7 touch the workflows and are a second PR.

moshehbenavraham:
Closes #178.

Compute mobile totals once after each complete prototype, run independent test targets concurrently, replace fixed test sleeps with polling, and add `CUTEST_FILTER` plus slow-test reporting. The complete development-world CuTest run fell from **35.7 s to 6.0 s**, retaining all **1,448 tests**.

CI reuses one build for both I/O drivers and the startup smoke, retains all five hardened builds with hardened tests on GCC 14/Autotools and Clang/CMake, splits archive checks, and caches compiler output. A reusable local container runner executes the workflow commands with isolated databases and shared ccache. Build/test guidance documents incremental builds and optional development flags. Shared calculator prerequisites prevent parallel compilation races; supervision fixtures also work while development port 4100 is occupied. Review follow-ups: the startup smoke test bounds every teardown wait and escalates to `SIGKILL` (with `timeout-minutes` on the job and a per-container timeout in the local runner), the authoritative cutest entry points clear `CUTEST_FILTER` and the runner reports a narrowed selection, and the setup guide documents the ccache `PATH` masquerade instead of `CC="ccache gcc"`.

Validation at `ec8fbf9a6` (rebased onto master `b7483d370`):

- All **20 local matrix jobs passed in 479.69 s**, including sanitizers, Valgrind, coverage, archive builds, and database migrations.
- Three `make -j16 test-all` runs passed in **12.6 / 11.5 / 11.6 s**. Three parallel CTest runs passed **28/28**, about **10.7 s** each.
- All **27,092** mobile prototype dumps compare byte-for-byte: **3,847,064 bytes** before and after.
- Supervision fixtures passed ten consecutive runs with port 4100 occupied. Filter and slow-failure regressions, dependency checks, manifest parity, hooks, and workflow syntax passed.
- The consolidated container job passed both I/O drivers (1,453 tests each), real-server startup on port 4100, health and graceful shutdown, installed-binary assertions, clean-tree checks, and distribution hygiene.
- The reviewer's SIGTERM-ignoring stand-in server now fails the smoke test in about 22 s with no leftover processes; a filtered run prints `CUTEST_FILTER=Test_mob_autoroll: 4 of 1453 tests selected`.

Validation uses the complete development world through an Ubuntu 24.04 toolchain container. The issue's stated build removals yield **21 executable builds including CodeQL**; the original estimate of 23 was an arithmetic error.

## Summary by CodeRabbit

- **New Features**
  - Added `CUTEST_FILTER` support for focused test runs, with clear selection and unmatched-filter feedback.
  - Added slow-test timing, highlighting tests exceeding one second.
  - Added a local Docker-based CI runner for running workflow jobs locally.
  - Added configurable autorun restart delays through `AUTORUN_FASTBOOT_DELAY`.
  - Added startup smoke testing for server and health endpoints.

- **Bug Fixes**
  - Corrected derived-stat recalculation for mobile prototypes.

- **Performance**
  - Enabled parallel test execution to shorten build and validation times.

- **Documentation**
  - Updated setup, testing, and development guidance for filtering, caching, parallel execution, and local CI.


scripts/ci/test_server_startup.sh
```diff
@@ -0,0 +1,79 @@
...
+server_log="$sandbox/server.log"
+server_pid=
+supervisor_pid=
+
+cleanup()
+{
+  if [[ -n "$supervisor_pid" ]]; then
+    (cd "$sandbox" && MUD_PORT=4100 ./scripts/autorun/autorun.sh stop) >/dev/null 2>&1 || true
+    kill "$server_pid" 2>/dev/null || true
+    wait "$supervisor_pid" 2>/dev/null || true
```
moshehbenavraham: **[High] Failure cleanup can hang the job for 6 hours (local matrix: forever) instead of failing**

**Location:** `cleanup()` lines 16-19, reached from every `fail`/`set -e` exit once the supervisor is running.

**Why it blocks:** this `wait` returns only when the autorun supervisor exits, and autorun exits only after the MUD exits:
- `autorun.sh` `handle_sigterm` deliberately ignores SIGTERM ("ignoring to keep autorun alive").
- `autorun.sh stop` sends the MUD one SIGTERM and never escalates (`signal_verified_process`).
- The server's SIGTERM handler (`hupsig`) only sets `shutdown_requested`; under libevent it is dispatched from the reactor. A server whose event loop or shutdown path stalls never exits.
- Nothing here sends SIGKILL, `unit-tests` has no `timeout-minutes`, and `scripts/ci/local/run.py` runs `docker run` with no timeout.

A server that does not finish shutting down after SIGTERM is exactly what this smoke test exists to catch. The removed `integration.yml` job reported it within about 6 s (TERM, 4 s poll, `::error::`, then TERM, 2 s, KILL). Here the step prints `Server did not exit successfully` and then hangs until GitHub's 360-minute job limit; the local container matrix never completes.

**Repro:** HEAD `scripts/ci/test_server_startup.sh`, `scripts/autorun/autorun.sh`, and `scripts/operations/healthcheck.sh` copied into a scratch repo `$R/repo`, plus `$R/runtime/etc/config`. `bin/luminari` is a stand-in (below) that listens on 4100, returns a healthy `/health` on 4182, and ignores SIGTERM, i.e. a stalled graceful shutdown. It runs in private net/pid namespaces, so no real 4100 listener is touched:

```sh
export R=/path/to/scratch
unshare -rn --pid --fork --mount-proc bash -c 'ip link set lo up
  exec unshare --map-user=1000 --map-group=1000 bash -c "TMPDIR=$R/tmp LUMINARI_TEST_DATA_DIR=$R/runtime timeout 45 $R/repo/scripts/ci/test_server_startup.sh; echo exit=\$?; ps -eo pid,args | grep -E \"autorun.sh|bin/fake/luminari\" | grep -v grep"'
```

Result:

```
Server is accepting connections on port 4100
{"service":"luminari-mud","status":"healthy","database":"healthy"}
Server did not exit successfully
exit=124        <- killed by timeout after 47 s, still blocked in cleanup's wait
   14 /bin/bash ./scripts/autorun/autorun.sh foreground          <- still running
  124 /usr/bin/python3 .../bin/fake/luminari -f lib/etc/config -d lib -o .../server.log -q -s 4100
```

<details><summary>Stand-in server (<code>bin/luminari -> fake/luminari</code>, outside <code>releases/</code> so autorun accepts it without a manifest)</summary>

```python
#!/usr/bin/python3
import http.server, signal, socket, sys, threading
args = sys.argv[1:]
if '-o' in args:
    with open(args[args.index('-o') + 1], 'a') as log:
        log.write('fake server booted\n')
signal.signal(signal.SIGTERM, signal.SIG_IGN)  # graceful shutdown never completes
class Health(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        body = b'{"service":"luminari-mud","status":"healthy","database":"healthy"}'
        self.send_response(200)
        self.send_header('Content-Length', str(len(body)))
        self.end_headers()
        self.wfile.write(body)
threading.Thread(target=http.server.HTTPServer(('127.0.0.1', 4182), Health).serve_forever,
                 daemon=True).start()
listener = socket.create_server(('0.0.0.0', 4100))
while True:
    listener.accept()[0].close()
```
</details>

**Fix:** bound every teardown wait and escalate:
1. Start the supervisor in its own process group (`setsid ... &`) and record its pgid.
2. After `autorun.sh stop`, both in `cleanup` and after the shutdown poll, poll up to ~10 s for `server_pid` to exit, then `kill -KILL "$server_pid"`.
3. Poll up to ~5 s for the supervisor, then `kill -KILL -- -"$pgid"`, and only then `wait`. Treat any escalation as a test failure.
4. Add `timeout-minutes` to the `unit-tests` job and `timeout=` to the `subprocess.run` of `docker run` in `run.py`, so any future unbounded wait fails fast.

Makefile.am
```diff
@@ -1043,17 +1047,21 @@ distcheck-archive:
...
 	test-binary-name-static test-healthcheck test-sql-interpolation test-production-profile \
 	test-background-help-entries test-demand-driven-architecture test-event-admission \
 	test-event-core-performance-gate test-native-event-architecture test-pubsub-retirement \
 	test-vessel-tooling
+
+# Build check programs once before running the suite, alongside the shell gates.
+run-cutest: check
 	LUMINARI_TEST_ROOT="$(abs_top_srcdir)" \
 		LUMINARI_TEST_SPEC_WORLD_ROOT="$(abs_top_srcdir)/unittests/CuTest/fixtures/spec_world_inventory" \
 		./cutest
```
moshehbenavraham, line 1065: **[Medium] An inherited `CUTEST_FILTER` silently narrows the authoritative gates**

**Location:** the `run-cutest` recipe (lines 1058-1061). The same applies to CTest `production-cutest` at `CMakeLists.txt:1147-1151`, whose `ENVIRONMENT` sets only the two root variables.

**Failure:** both entry points pass the caller's environment straight to `./cutest`. This PR documents `CUTEST_FILTER` for focused iteration (AGENTS.md, TESTING_GUIDE.md). A value left exported in the shell makes `make -j test`, `make -j test-all`, and `ctest --preset ...` report success after running a handful of tests. Those are the commands the same docs name as full validation. "Full validation must leave the filter unset" is written down but not enforced.

**Repro** (built HEAD `cutest`, exact `run-cutest` command line):

```
$ CUTEST_FILTER=Test_mob_autoroll LUMINARI_TEST_ROOT="$PWD" \
    LUMINARI_TEST_SPEC_WORLD_ROOT="$PWD/unittests/CuTest/fixtures/spec_world_inventory" ./cutest
....

OK (4 tests)
$ echo $?
0
```

4 of 1,448 tests ran. The shell gates are unaffected and pass, so `make -j test-all` finishes green and installs.

**Fix:** pin the filter off in the authoritative paths: `env -u CUTEST_FILTER ./cutest` in `run-cutest` (or `CUTEST_FILTER= ./cutest`, since empty already means all tests), and add `CUTEST_FILTER=` to `production-cutest`'s `ENVIRONMENT`. Also have the generated runner print `CUTEST_FILTER=<value>: N of M tests selected` whenever a filter is active, so a narrowed run is visible in any log.

docs/guides/SETUP_AND_BUILD_GUIDE.md
```diff
@@ -36,19 +36,35 @@ require the same libraries.
...
+Use `make clean` after changing compiler flags or build configuration, or when dependency
+files are stale; it is not part of the normal edit/test loop.
+
+Optional compiler caching and faster unoptimized development builds:
+
+```bash
+sudo apt install ccache
+./configure CC="ccache gcc"
+# Optional: use this instead when optimizing edit/compile latency.
+./configure CC="ccache gcc" CFLAGS="-g -O0"
```
moshehbenavraham: **[Medium] Following this ccache recipe makes `make test` fail**

**Location:** lines 52-54 (`./configure CC="ccache gcc"`, including the `-O0` variant).

**Failure:** configure records `CC = ccache gcc`, and `make test` passes it on as `CC="$(CC)"` to `test-production-profile` and `test-native-event-architecture`. Both scripts execute it as a single word: `"$cc"` in `test_production_profile.sh` and `production_profile.sh:134,139`, and `"${CC:-cc}"` in `test_native_event_architecture.sh:53-77`. The shell looks for a program literally named `ccache gcc`. `./configure --enable-production` breaks the same way through `production_profile.sh --cc "$CC"`.

**Repro** (a `ccache` shim that only does `exec "$@"`, so just the documented `CC` value is exercised):

```
$ PATH=$shim:$PATH CC="ccache gcc" scripts/deployment/test_production_profile.sh
production profile: ccache gcc cannot build a trivial program: .../production_profile.sh: line 139: ccache gcc: command not found
production profile test: production probe failed for ccache gcc                 (exit 1)
$ PATH=$shim:$PATH CC="ccache gcc" scripts/events/test_native_event_architecture.sh
.../test_native_event_architecture.sh: line 53: ccache gcc: command not found  (exit 127)

$ PATH=$shim:$PATH CC=gcc scripts/deployment/test_production_profile.sh         (control)
production profile test: PASS (gcc: fortify-source pie relro bind-now ...)      (exit 0)
$ PATH=$shim:$PATH CC=gcc scripts/events/test_native_event_architecture.sh
native event architecture test: PASS                                            (exit 0)
```

**Fix:** document the masquerade form the new CI action already uses: `export PATH=/usr/lib/ccache:$PATH` with a plain `./configure` (and `./configure CFLAGS="-g -O0"` for the unoptimized variant). Alternatively, make both scripts split `CC` into an argv array (`read -r -a cc_cmd <<< "$CC"`) before invoking it.

moshehbenavraham (commented):
Adversarial review of `1dba258fe`. Three defects, each with a reproduction and a fix, are inline:

1. **High** - `scripts/ci/test_server_startup.sh` cleanup blocks forever when the server does not exit on SIGTERM: `unit-tests` hangs to the 360-minute job limit (the local container matrix never finishes) instead of failing in seconds like the removed integration job.
2. **Medium** - an inherited `CUTEST_FILTER` makes `make test`, `make test-all`, and CTest `production-cutest` pass after running 4 of 1,448 tests.
3. **Medium** - the new `./configure CC="ccache gcc"` guidance breaks `make test` (production-profile and native-event-architecture gates).

Checked and found sound: the `db.c` `affect_total` move (espec handlers assign base fields only; `affect_total_sub` rebuilds `aff_abils`/`points` from base values; for NPCs `compute_char_cap` only clamps `points.size` and the CON delta is zero without affects/gear), `check: rol-mob-calculator` ordering (it is a prerequisite of Automake's `check` recipe), rename-schema DB isolation (private socket-only mariadbd), the select-driver `./cutest` run without `LUMINARI_TEST_ROOT` (all consumers fall back to `.`), and the supervision `ss` shim (fixtures never listen, so semantics are unchanged).

scripts/ci/test_server_startup.sh
moshehbenavraham: Fixed. `scripts/ci/test_server_startup.sh` now starts the supervisor with `setsid` so it and the server own a separate process group, and every teardown wait is bounded: after `autorun.sh stop` it polls up to 10 s for the server and sends `SIGKILL` if it is still alive, then polls up to 5 s for the supervisor and kills the whole group before `wait`. The success path uses the same bounded wait, so a supervisor that logs "terminated gracefully" but never exits also fails instead of hanging. The `unit-tests` job has `timeout-minutes: 45`, and the local runner passes `timeout=` to `docker run` (default 45 minutes, `--timeout` overrides) and kills the named container on expiry so a stuck job cannot outlive the client.

Reproduced with the stand-in server from this thread in private net/pid namespaces: the script now prints `Server ignored SIGTERM; killing it`, exits 1 after about 22 s, and leaves no `autorun.sh` or fake server process behind.

Makefile.am
moshehbenavraham: Fixed. `run-cutest` now runs `CUTEST_FILTER= ... ./cutest`, and the CTest `production-cutest` environment sets `CUTEST_FILTER=`, so an exported filter cannot narrow `make test`, `make test-all`, or `ctest`. The generated runner prints `CUTEST_FILTER=<value>: N of M tests selected` whenever a filter is active; `scripts/ci/test_cutest_runner.py` asserts the line for a filtered run and its absence for full and empty-filter runs. AGENTS.md and the testing guide describe the behaviour.

Verified: `CUTEST_FILTER=Test_mob_autoroll make -n run-cutest` shows the cleared variable, and the real binary reports `CUTEST_FILTER=Test_mob_autoroll: 4 of 1453 tests selected`.

docs/guides/SETUP_AND_BUILD_GUIDE.md
moshehbenavraham: Fixed in the docs. The setup guide now shows the masquerade form that CI already uses (`export PATH=/usr/lib/ccache:$PATH` with a plain `./configure`, or `./configure CFLAGS="-g -O0"`) and states explicitly that `CC="ccache gcc"` breaks the production-profile and native-event-architecture gates. The same note went into `docs/ongoing-projects/test-loop-speed.md`. I left the shell gates running `$CC` as a single word, since the CI action and the documented recipe both keep `CC=gcc`.

moshehbenavraham:
Rebased onto master (`b7483d370`) and addressed all three review findings in `ec8fbf9a6`; each thread has the fix and its reproduction. All 20 workflow jobs were run locally through the container matrix on that commit and passed (479.69 s), including the consolidated unit-tests job with the real-server startup smoke test. Ready to squash-merge.

PR #193, opened 2026-09-16
Decompose three oversized translation units and rank the rest

### LINKED ISSUE #96: Modernization: decompose oversized translation units and strengthen internal C API boundaries
## Why

The source tree already has sensible first-level subsystem directories, but many translation units remain too large to analyze, test, compile incrementally, or modernize safely as coherent units.

- Approximately 125 tracked C files exceed 1,000 lines and 25 exceed 5,000 lines.
- Examples include `src/character/perks.c` (about 23,588 lines), `src/combat/fight.c` (17,279), `src/combat/act.offensive.c` (16,506), `src/magic/magic.c` (15,559), `src/act.wizard.c` (13,558), `src/act.other.c` (13,531), and `src/utils.c` (12,250).
- Size alone is not a defect: generated data, tables, unavoidable literals, and cohesive dispatch code need different treatment. The risk is mixed responsibilities, broad mutable state, weakly owned interfaces, and changes whose test surface is difficult to isolate.
- Header and source boundaries currently expose more internal state than necessary, increasing warning cleanup and concurrency risk.

## Scope

- Produce module metrics that exclude comments/generated tables and combine size with churn, fan-in/fan-out, global-state access, compile time, and defect/security criticality.
- Prioritize extraction by responsibility and risk, not a blanket line limit.
- Begin with a few high-value seams such as combat resolution vs presentation, wizard command families, parsing vs execution, and generated/static perk/spell data vs algorithms.
- Preserve the repository's one-level `src/<subsystem>/` layout; do not add a second directory level.
- Introduce internal headers only where multiple files share a contract, path-qualifying cross-subsystem includes as required by repository convention.
- Make file-private functions/state `static`; use opaque types or accessor APIs for ownership-sensitive structures where practical.
- Document ownership, borrowing, mutation, nullability, lifetime, thread-safety, and error behavior for subsystem APIs.
- Reduce cross-subsystem globals and route dependencies through narrow interfaces.
- Add characterization tests before extraction and verify behavior after each small move.
- Update both `Makefile.am` and `CMakeLists.txt` for every source split and keep their manifests in parity.
- Identify large generated tables and move their generation/input contract out of hand-maintained algorithmic code rather than mechanically splitting them.

## Acceptance criteria

- A checked-in report ranks initial candidates by risk/churn/coupling and labels generated/table-heavy exceptions.
- The first modernization wave extracts at least three coherent boundaries with no behavior change.
- Extracted modules have focused tests and documented API ownership/lifetime rules.
- New internal symbols default to file scope; public headers expose only required contracts.
- Autotools/CMake source lists and tests remain synchronized.
- Compile time, dependency fan-out, or changed lines per responsibility measurably improve for each extraction.
- No mechanical restyling or unrelated rewrite is mixed into the structural changes.

moshehbenavraham:
Closes #96.

Adds module metrics plus a checked-in ranking so decomposition targets are chosen by risk rather
than line count, then extracts the first three seams that ranking selected. The decomposition
commit is a pure move with no behaviour changes. Review then found five pre-existing defects in the
moved combat code. They are fixed in a separate commit so the move stays mechanically verifiable;
see *Review fixes* below.

## The issue is still valid

The audit in #96 was taken at `d52d2dc6`. The files have grown since: `perks.c` 23,588 -> 23,614,
`fight.c` 17,279 -> 18,153. Only the paths changed, when sources moved under `src/<subsystem>/`.

## Metrics and report

`scripts/development/module_metrics.py` measures size, non-comment code, initialiser-table share,
churn, fix churn, fan-in, fan-out and external symbol count per translation unit, normalises each
against the tree, and ranks the result. Churn follows renames, so the `src/<subsystem>/` move does
not read as "every file was rewritten last month". Table-heavy files are labelled as exceptions
rather than ranked as candidates, because size alone is not a defect.

`docs/development/OVERSIZED_TRANSLATION_UNITS.md` carries the ranking taken at the branch point,
the table-heavy exception list, why these three seams came first, the measured effect, and a
suggested next wave.

The pre-wave ranking put `fight.c` at 3, `perks.c` at 5 and `act.wizard.c` at 8. Ranks 1, 2, 4 and
6 (`db.c`, `utils.c`, `interpreter.c`, `comm.c`) were deliberately left alone: their fan-in is 240
to 403 files, so touching those interfaces needs its own wave.

## The three extractions

| Seam | From | To |
| -- | -- | -- |
| Perk tables vs perk engine | `src/character/perks.c` | `src/character/perk_definitions.{c,h}` |
| Combat presentation vs resolution | `src/combat/fight.c` | `src/combat/combat_messages.{c,h}` |
| The staff `set` command family | `src/act/act.wizard.c` | `src/act/act.wizard.set.c` |

**Perk tables.** The fifteen `define_*_perks()` functions are 9,994 lines of table written as field
assignments, with no algorithmic content. #96 asks for exactly this treatment and says to move the
contract out *rather than* mechanically splitting, so it is one data file. `perks.h` no longer
declares them; `define_druid_perks()`, declared but never defined anywhere, was dropped.

**Combat presentation.** `dam_message()`, `skill_message()` and their helpers leave `fight.c`, so
retuning combat text cannot change a fight. `skill_message()` had no caller outside `fight.c`, so
its declaration leaves the widely included `fight.h` for the internal `combat_messages.h` - a net
reduction in public surface, not a lateral move.

**The `set` family.** Needed no new header at all: `do_set` is reached through the command table
and the test hook already lives in `act/act.h`. `set_fields[]` had external linkage with no
external readers, so the move made it `static`, deleting a global.

Each new header documents ownership, lifetime, nullability, call order and thread safety.
`get_wielded()` gains the prototype it never had outside `fight.c`.

## Measured effect

Best of seven round-robin runs with the project's own flags, reported as CPU time because other
work was running on the host.

| Editing this responsibility | Before | After | Change |
| -- | --: | --: | --: |
| A class's perk table | 3.99s | 1.53s | -62% |
| The perk engine | 3.99s | 1.99s | -50% |
| Combat text | 3.00s | 0.14s | -95% |
| Combat resolution | 3.00s | 2.84s | -5% |
| The `set` command | 2.63s | 0.27s | -90% |
| Other staff commands | 2.63s | 2.34s | -11% |

Header fan-out for the extracted responsibility: perk tables 53 -> 30, combat text 98 -> 33, the
`set` command 106 -> 65. Full-build cost is flat, as expected; the gain is incremental work and
blast radius.

## Defects the new tests surfaced

Characterisation tests on the perk tables found seven pre-existing content defects. None are fixed
here: correcting any changes what players can see or buy, which is a content decision that needs
its own review. They are recorded as ratchets, so the tables cannot get worse and can be fixed
freely.

- `PERK_WIZARD_EXTENDED_SPELL_3` requires `PERK_WIZARD_EXTENDED_SPELL_2`, which has an id in
  `structs.h` but is never defined, so the perk is permanently unbuyable.
- `PERK_CLERIC_DOMAIN_FOCUS_3`, `PERK_CLERIC_DIVINE_SPELL_POWER_3` and
  `PERK_CLERIC_GREATER_TURNING` each require rank 5 of a prerequisite whose `max_rank` is 2 or 3.
- `PERK_BARBARIAN_RAGE_ENHANCEMENT`, `PERK_BARBARIAN_EXTENDED_RAGE_1` and
  `PERK_BARBARIAN_TOUGHNESS` never set `perk_category`, so they appear under no tree.

They are tracked in #194. Writing them up also turned up #195: `perk_category_names[]` has drifted
from the `PERK_CATEGORY_*` constants, so 320 of 562 perks show the wrong tree name.

## Review fixes

Adversarial review found five defects where `combat_messages.h`, now the contract callers read
instead of `fight.c`, did not match the code moved under it. All five predate this PR. They are
fixed in 6ec215c7c, and `docs/development/OVERSIZED_TRANSLATION_UNITS.md` records them
(752636c45).

| Finding | Fix | Pinned by |
| -- | -- | -- |
| Trelux stand-in object from `read_object()` never extracted: one leaked object per combat message | Single exit in `skill_message_with_projectile()` that extracts it | `..._release_the_trelux_stand_in_object` |
| Test fixture asserted before teardown; CuTest's `longjmp` left `world` dangling | Record, tear down, then assert, in every test | Injected-failure probe: all globals restored |
| `dam_message()` indexed `attack_hit_text[]` unbounded; `IS_WEAPON()` was private to `fight.c` | `IS_WEAPON()` moved to `structs.h`; `dam_message()` rejects bad types with a SYSERR; `replace_string()` bounded | `..._dam_message_rejects_a_non_weapon_type` (segfaults without the guard) |
| Header said `dam_message()` always renders; at `POS_DEAD` it renders nothing | Header corrected; the corpse fallback in `damage_with_projectile()` now logs a SYSERR instead of going quiet | `..._dam_message_is_silent_for_a_dead_victim` |
| `armor_list[]` indexed by `value[1]` of non-armour body items | Guarded `GET_ARMOR_TYPE_PROF()` | Reviewed against world data |

Each new test was checked to fail with its fix reverted.

One of these fixes is visible to players. Four wand-type crests that mobs in zone 20202 wear on the
body, plus Graye's Staff (vnum 31020), have a charge count in `value[1]` that selected an armour
row.
Misses against their wearers could read "glances off" the item; they now get the ordinary miss
line. No spec proc is bound to those objects, so the change is text only. The other four fixes do
not change anything players see with current data.

## Verification

- `make -j$(nproc)` and `make install` clean: zero warnings under the full strict tier
  (`-Wconversion -Wshadow -Wjump-misses-init -Wcast-align=strict ...`).
- Full CuTest suite: **1503 tests, OK**, run with the harness environment
  (`LUMINARI_TEST_ROOT`, `LUMINARI_TEST_SPEC_WORLD_ROOT`).
- `test_combat_messages.c` under valgrind (`--leak-check=full --track-origins=yes
  --track-fds=yes`): 0 errors, 0 bytes definitely lost, no leaked descriptors.
- Local CI with `scripts/ci/local/run.py`: all 28 jobs from `test`, `integration`, `quality`,
  `hygiene` and `security` pass on 6ec215c7c. The later 752636c45 only touches docs, which
  triggers just `quality` and `hygiene`; both pass on it. On GitHub, the two `CMake gcc-16` jobs
  and the `CodeQL` check fail on master in exactly the same way; details are in the review-round
  comment.
- `python3 scripts/ci/check_build_parity.py` passes; both manifests updated for all four new
  sources and three new tests.
- `pre-commit run --files ...` clean over every changed file.
- Pre-push build verification passed.

Two suite failures seen early on were a worktree setup gap, not a regression: this worktree had no
generated `lib/world/*/index` and no world data, and `./cutest` was being run without the
harness's environment variables. Both were fixed and the tests then passed.

## Review notes

Review it commit by commit:

1. 0e3fe4ff6, the decomposition. `git show --stat` is the fastest way in. `perks.c` and `fight.c`
   gain exactly one line each (an `#include`), `act.wizard.c` gains none, and everything else in
   the parents is deletion. The moved blocks are
   byte-for-byte identical apart from these deliberate changes: `static` removed from
   `dam_message`, `skill_message_with_projectile` and `define_wizard_controller_perks` (their
   internal headers now declare them); `static` added to `set_fields[]`; and three new test hooks
   in the existing `#ifdef LUMINARI_CUTEST` block of `act.wizard.set.c`
   (`find_set_field_for_test`, `set_field_count_for_test`, `set_field_min_level_for_test`,
   declared in `act/act.h`).
2. a8d0898bb links the perk defects to #194 and #195 in the report.
3. 6ec215c7c holds the review fixes above. It is the only commit that changes behaviour.
4. 752636c45 records those fixes in the report.

Please merge with **Rebase and merge** rather than squash. Squashing would fold a 13,000-line pure
move and the behaviour fixes into one commit, so neither could be checked on its own and a
`git bisect` would land on the whole thing.


src/combat/combat_messages.c
```diff
@@ -0,0 +1,703 @@
...
+    weap = get_wielded(ch, dualing); /* four arms: the lower-arm weapon */
+  else if (GET_EQ(ch, WEAR_WIELD_2H))
+    weap = GET_EQ(ch, WEAR_WIELD_2H);
+  else if (dualing == 1)
+    weap = GET_EQ(ch, WEAR_WIELD_OFFHAND);
+
+  /* special handling for Trelux */
+  if (GET_RACE(ch) == RACE_TRELUX)
+  {
+    weap = read_object(TRELUX_CLAWS, VIRTUAL);
```
moshehbenavraham: ## 1. `read_object()` here is never extracted - unbounded object leak, and the new header states the opposite

`combat_messages.h:24-25` is the contract this PR writes for this module:

> The one allocation is internal: `skill_message_with_projectile()` may `read_object()` a stand-in claw object to name a natural attack, **and it extracts that object before returning. Callers see no allocation.**

That is not what the code does. `grep -c extract_obj src/combat/combat_messages.c` is **0**. There is no free, no `extract_obj`, no `obj_from_char` anywhere in the file, and `weap` is overwritten two lines later on the ranged path (`:383`) and the shield path (`:408`), so on those paths the pointer is lost before any cleanup could reach it.

`read_object()` (`src/core/db.c:5114`) is an allocator with global side effects:

```c
CREATE(obj, struct obj_data, 1);
clear_object(obj);
*obj = obj_proto[i];
obj->next = object_list;      /* linked into the global list        */
object_list = obj;
obj->object_list_member = true;
add_obj_to_rnum_hash(obj);    /* and into the rnum hash             */
obj_index[i].number++;        /* and counted as "in the world"      */
```

**Repro:** play a Trelux (registered as a full PC race at `src/character/race.c:1657`; vnum 800 exists at `lib/world/obj/8.obj:1`). Every call into `skill_message_with_projectile()` with a Trelux attacker allocates one `obj_data`. `damage()` reaches it on every miss, every death blow, and every skill/spell damage event (`src/combat/fight.c:6057`, `:6078`). One leak per event, for the life of the boot.

**Impact, in order of how much it hurts:**

1. `obj_index[real_object(800)].number` grows without bound. Any zone-reset or spec-proc condition of the form "load only if fewer than N exist" for that vnum is permanently poisoned.
2. `object_list` grows without bound, and it is a singly linked list that several subsystems walk in full. Every traversal gets slower for the rest of the boot.
3. Straight memory growth, proportional to Trelux combat volume.

The objects are also in `object_list` while being in no room and in nobody's inventory, which is a state some `object_list` walkers do not expect.

**Fix.** The only thing `weap` is used for is `act()`'s `$p` argument, so the cheapest correct change is to stop allocating: resolve the short description from `obj_proto[real_object(TRELUX_CLAWS)]` and pass `NULL` with a literal noun. If the live object is wanted, the function needs a single exit:

```c
struct obj_data *trelux_claws = NULL;
...
if (GET_RACE(ch) == RACE_TRELUX)
{
  trelux_claws = read_object(TRELUX_CLAWS, VIRTUAL);
  weap = trelux_claws;          /* may be NULL; act() tolerates it */
  attacktype = TYPE_CLAW;
}
...
/* every `return X;` becomes `{ return_value = X; goto done; }` */
done:
  if (trelux_claws)
    extract_obj(trelux_claws);
  return (return_value);
```

Note there are four returns to route (`:456`, `:474`, `:512`, `:695`) plus the fall-through at `:699`.

The leak predates this PR - it moved verbatim out of `fight.c`. What is new is `combat_messages.h` telling every future reader that it has been handled. Either fix the code or fix lines 22-26 of the header; leaving both is worse than the original, because the original did not make a promise.

unittests/CuTest/test_combat_messages.c
```diff
@@ -0,0 +1,231 @@
...
+
+  /* dam_message() is the fallback of last resort: unlike skill_message() it
+   * has no "no message" answer, so both participants must always be told
+   * something. */
+  dam_message(10, &fix.attacker, &fix.victim, TYPE_HIT, 0, NULL);
+
+  CuAssertTrue(tc, strlen(fix.attacker_desc.small_outbuf) > 0);
+  CuAssertTrue(tc, strlen(fix.victim_desc.small_outbuf) > 0);
+
+  combat_messages_teardown(&fix);
```
moshehbenavraham, line 318: ## 2. Every test in this file restores `world` only in teardown, and CuTest `longjmp`s past teardown on a failed assert

`CuTest.c:179-189`:

```c
static void CuFailInternal(CuTest *tc, const char *file, int line, CuString *string)
{
  ...
  if (tc->jumpBuf != 0)
    longjmp(*(tc->jumpBuf), 0);
}
```

Every `CuAssert*` in this file is placed **before** its `combat_messages_teardown(&fix)` call - `:136`/`:138`, `:154`/`:156`, `:172-173`/`:175`, `:192`+`:196-198`/`:200`, `:224-226`/`:228`. So the first assertion that ever fails skips teardown, and `setup()` (`:89-97`) has already done this:

```c
fix->saved_world = world;
fix->saved_top_of_world = top_of_world;
fix->saved_character_list = character_list;
world = &fix->room;          /* &fix->room is a stack local */
top_of_world = 0;
character_list = NULL;
```

`world` is left pointing into the dead frame of a function that has been `longjmp`ed out of, permanently, for the rest of the process. The next test's `setup()` then *saves that stale pointer* as its "original", so its teardown restores the dangling pointer rather than the real one. The corruption is self-propagating and there is no path back.

**Measured.** I flipped one assertion at `:172` to fail (`strlen(...) > 999999`), added a probe test immediately after this file's tests, and ran the full suite both ways:

```
clean run:            [PROBE] world=(nil)            top_of_world=0 character_list=(nil)   OK (1501 tests)
one failed assert:    [PROBE] world=0x7fff2d4dee60   top_of_world=0 character_list=(nil)   Runs: 1501 Passes: 1500 Fails: 1
```

`0x7fff...` is the stack. The tree is back to its original state; the probe and the injected failure are not in the branch.

**Why this is worth fixing even though the run above still reported cleanly.** The only reason the dangling pointer is not dereferenced today is scheduling: these tests happen to land before the world is booted, so the "real" `world` they clobber is `NULL` anyway. Test order here is data - the position of `test_combat_messages.c` in `cutest_test_files`. Move this file down the list, or boot the world earlier, and one honest assertion failure silently hands a freed stack pointer to roughly a thousand downstream tests. That failure mode does not present as "test 3 failed"; it presents as a segfault somewhere unrelated, which is a bad afternoon.

**Fix** - the pattern this PR already edits two files over, `test_spec_combat_secondary.c:487-524`: record into `bool`s, tear down, then assert.

```c
void Test_combat_messages_dam_message_always_produces_text(CuTest *tc)
{
  struct combat_messages_fixture fix;
  bool attacker_told;
  bool victim_told;

  combat_messages_setup(&fix);
  combat_messages_reset_output(&fix.attacker_desc);
  combat_messages_reset_output(&fix.victim_desc);

  dam_message(10, &fix.attacker, &fix.victim, TYPE_HIT, 0, NULL);

  attacker_told = strlen(fix.attacker_desc.small_outbuf) > 0;
  victim_told = strlen(fix.victim_desc.small_outbuf) > 0;

  combat_messages_teardown(&fix);

  CuAssertTrue(tc, attacker_told);
  CuAssertTrue(tc, victim_told);
}
```

`Test_combat_messages_substitute_the_weapon_noun` needs the same treatment for its `CuAssertPtrNotNull(tc, singular)` at `:192`, which sits between setup and the work and would strand the globals before the test has done anything at all.

src/combat/combat_messages.c
```diff
@@ -0,0 +1,703 @@
...
+      {"*THFFFT* $n fires $p at $N *THWAK* lethally wounding $M!", /* 12: dam <= 50% */
+       "\t[f500]*THFFFT*\tn \t[f030]you fire \tn$p\tn \t[f030]at \tn$N\tn \t[f030]*THWAK* lethally "
+       "wounding $M!\tn",
+       "*THFFFT* $n fires $p at you *THWAK* lethally wounding you!"},
+      {"*THFFFT* $n fires $p at $N *THWAK* nearly killing $M!", /* (13): > 51   */
+       "\t[f500]*THFFFT*\tn \t[f030]you fire \tn$p\tn \t[f030]at \tn$N\tn \t[f030]*THWAK* nearly "
+       "killing $M!\tn",
+       "*THFFFT* $n fires $p at you *THWAK* nearly killing you!"}};
+
+  w_type -= TYPE_HIT; /* Change to base of table with text */
```
moshehbenavraham, line 205: ## 3. `dam_message()` became public, but the only thing that bounded `w_type` stayed private to `fight.c`

`:192` rebases the caller's `w_type` onto a 24-entry table and then indexes it three times unchecked (`:282-283`, `:298-299`, `:314-315`):

```c
w_type -= TYPE_HIT;
...
buf = replace_string(dam_weapons[msgnum].to_room,
                     attack_hit_text[w_type].singular,
                     attack_hit_text[w_type].plural);
```

`attack_hit_text[]` is defined at `src/combat/fight.c:97` with exactly 24 entries (`TYPE_HIT` 2300 .. `TYPE_GORE` 2323). The single thing that has ever kept `w_type` inside that range is the caller's gate:

```c
/* src/combat/fight.c:163 */
#define IS_WEAPON(type) (((type) >= TOP_ATTACK_TYPES) && ((type) < BOT_WEAPON_TYPES))
```

That macro is `#define`d **inside `fight.c`**. It is not in a header, so no new caller of the now-public `dam_message()` can even see it, and `combat_messages.h:60-66` - which is otherwise meticulous, documenting ownership, borrowing, nullability, buffer lifetime and thread safety - says nothing about `w_type` at all. The one caller the refactor left behind still gates correctly (all three call sites at `fight.c:6076`, `:6081`, `:6088` are inside the `else` of `if (!IS_WEAPON(w_type))`), so nothing is broken today. The refactor simply deleted the coupling that made that true and did not replace it with anything.

**What a violation costs.** `attack_hit_text[w_type]` out of range yields two garbage `const char *` values, which go straight into `replace_string()` at `:37`:

```c
static char *replace_string(const char *str, const char *weapon_singular, const char *weapon_plural)
{
  static char buf[MEDIUM_STRING] = {'\0'};   /* 256 */
  char *cp = buf;
  ...
      case 'W':
        for (; *weapon_plural; *(cp++) = *(weapon_plural++))
          ;
```

No bound on `cp`, no bound on the source. A wild `weapon_plural` is both an unbounded read and an unbounded write into a 256-byte BSS buffer. Today it is safe only because both operands are string literals from a fixed table.

**Fix**, either or both:

1. Make the precondition enforceable rather than documented. Move `IS_WEAPON` into `structs.h` next to `BOT_WEAPON_TYPES`/`TOP_ATTACK_TYPES` (which already live there at `:5415-5416`), state the precondition in `combat_messages.h`, and reject at the top of `dam_message()`:

```c
if (!IS_WEAPON(w_type))
{
  log("SYSERR: dam_message: w_type %d out of range", w_type);
  return;
}
```

2. Give `replace_string()` the bound it has never had - track remaining space against `buf + sizeof(buf) - 1` in both copy loops. It is cheap and it turns a memory-safety bug into a truncated sentence.

The new test only ever passes `TYPE_HIT`, so it will not notice either way.

src/combat/combat_messages.c
```diff
@@ -0,0 +1,703 @@
...
+    {
+      buf = replace_string(dam_weapons[msgnum].to_victim, attack_hit_text[w_type].singular,
+                           attack_hit_text[w_type].plural);
+      GUI_CMBT_OPEN(victim);
+      act(buf, FALSE, ch, NULL, victim, TO_VICT | TO_SLEEP);
+      send_to_char(victim, CCNRM(victim, C_CMP));
+      GUI_CMBT_CLOSE(victim);
+    }
+  }
+  else
```
moshehbenavraham, line 335: ## 4. `dam_message()` is documented as "always produces output". At `POS_DEAD` it produces none - which is the exact case `fight.c` falls back to it for

The function is:

```c
if (is_ranged_weapon_attack(attack_type) && projectile && GET_POS(victim) > POS_DEAD)  { ... }
else if (GET_POS(victim) > POS_DEAD)                                                   { ... }
else
{
}                     /* <- :322-324 */
```

`POS_DEAD` is `0` (`structs.h:1043`) and it is the lowest position, so the empty branch is reached on exactly one condition: **the victim was killed by this blow**.

`combat_messages.h:63-66` states the opposite without qualification:

> Generic weapon damage text, scaled by the fraction of the victim's health removed. This is the fallback used when no skill message matches, and **it always produces output**.

And `fight.c` is built on that promise (`:6072-6082`):

```c
if (GET_POS(victim) == POS_DEAD || dam == 0)
{
  if (!dam && is_ranged)
    dam_message(dam, ch, victim, w_type, attack_type, projectile);
  else if (!skill_message_with_projectile(dam, ch, victim, w_type, attack_type, projectile))
  {
    /* no skill_message? try dam_message */
    dam_message(dam, ch, victim, w_type, attack_type, projectile);
  }
```

So on a killing blow whose attack type has no authored message, `damage()` hands off to `dam_message()` as the last resort and `dam_message()` emits nothing - no line to the killer, the victim, or the room, and no `SYSERR`.

**Reachability today: data, not code.** I checked `lib/misc/messages`; all 24 weapon types 2300-2323 have `M` blocks, so `skill_message_with_projectile()` currently always matches and returns `SKILL_MESSAGE_DEATH_BLOW` before the fallback fires. The guard is a text file with 340 entries against `MAX_MESSAGES 512`. Add a 25th `TYPE_*` (the comment at `structs.h:5410` - "don't forget to add to race_list() and all the other places if changed" - anticipates exactly this), or fat-finger one `M` block, and killing blows with that weapon go completely silent with nothing in the log to explain it.

The new characterisation test would not catch it: `Test_combat_messages_dam_message_always_produces_text` sets `GET_POS = POS_STANDING` and `GET_HIT = 100` and never exercises the dead branch, so it pins the invariant only where the invariant holds.

**Fix.** Pick one and make the header agree with it:

- Make it true: in the empty `else`, emit a minimal death line (the `msgnum == 0` / `dam_weapons[13]` text with the victim collapsing reads fine here) so the fallback is a real fallback.
- Or make it honest: change `combat_messages.h` to say `dam_message()` is silent at `POS_DEAD`, and have `fight.c:6078-6082` emit its own death line when `skill_message_with_projectile()` returns `SKILL_MESSAGE_MISS_FAIL` on a corpse.

Either way `log("SYSERR: no combat message for w_type %d", w_type + TYPE_HIT)` on that branch turns a silent content gap into something someone can act on. And the test deserves a `GET_POS(&fix.victim) = POS_DEAD` case so the invariant is pinned where it actually bites.

moshehbenavraham (commented):
# Adversarial review

Everything below assumes the PR is wrong until shown otherwise. Here is what survived that, so the findings are the only things I am asserting are broken.

## Independently verified

- **The three extractions really are pure moves.** I did not eyeball them: I extracted the full removed-line set of each parent and the full added-line set of each child and diffed the two sequences. They match line-for-line. The complete set of deltas is `static` removed from `dam_message`, `skill_message_with_projectile` and `define_wizard_controller_perks`; `static` added to `set_fields[]`; the deleted forward declarations; the new file headers; and three new `#ifdef LUMINARI_CUTEST` hooks in `act.wizard.set.c` (`find_set_field_for_test`, `set_field_count_for_test`, `set_field_min_level_for_test`). That last item is not covered by the PR body's "byte-for-byte identical apart from three deliberate changes" sentence - worth adding, since that sentence is what a reviewer will lean on to skip 13k lines.
- **Build is clean at the real warning tier.** `make -j` and `make cutest` both produce **0 warnings** under `-Wconversion -Wcast-qual -Wshadow -Wmissing-prototypes -Wfloat-equal -Wjump-misses-init -Wcast-align=strict -Wduplicated-branches -Wlogical-op ...`. No implicit declarations hiding in the new include sets.
- **Full suite reproduces: `OK (1500 tests)`**, run with the harness environment. `scripts/ci/check_build_parity.py` passes; both manifests carry all four new sources and all three new tests.
- **The perk tables are structurally sound.** 565 slot writes, 565 distinct slots, zero duplicates, every `perk->id` matches its own `perk_list[]` index, no two `PERK_*` names alias one numeric slot, and every index is inside `[0, NUM_PERKS)`. So no silent overwrite, no leaked `strdup()` from a double-written slot, no out-of-bounds write into the global at boot. This was my main worry about moving 10k lines of table and it is clean.
- **The claimed dead symbols really are dead.** `skill_message()` has no caller outside the combat pair, `define_druid_perks` has no reference anywhere in the tree, `set_fields[]` has no external reader. Narrowing all three is correct.
- **The new tests are memory-clean.** All three files under `valgrind --leak-check=full --track-origins=yes --errors-for-leak-kinds=definite`: 0 errors, 0 definitely lost.
- `Test_wizard_set_gates_every_field_behind_staff_level` is a better test than it looks. `show_set_help()` indexes a 4-entry `set_levels[]` by `level - LVL_IMMORT`, and `[LVL_BUILDER, LVL_IMPL]` is exactly `[31, 34]` - so that test is the out-of-bounds guard for `set help`, whether or not it was written as one.

## Findings

Five, all in `src/combat/combat_messages.c` and its header except #2. Details are inline on the relevant lines.

| # | Where | What |
| -- | -- | -- |
| 1 | `combat_messages.c:372`, `combat_messages.h:24` | `read_object()` is never extracted. Unbounded object leak per Trelux combat event, and the new header states it is cleaned up. |
| 2 | `test_combat_messages.c:175` | Every test restores `world` only in teardown; CuTest `longjmp`s past teardown on a failed assert, leaving `world` pointing into a dead stack frame permanently. Measured. |
| 3 | `combat_messages.c:192` | `dam_message()` went public; the `IS_WEAPON` bound that protected its unchecked `attack_hit_text[w_type]` stayed `#define`d inside `fight.c`. |
| 4 | `combat_messages.c:322` | Header says `dam_message()` "always produces output". At `POS_DEAD` it produces none - the exact case `fight.c:6081` falls back to it for. |
| 5 | `combat_messages.c:615` | `armor_list[GET_OBJ_VAL(armor, 1)]` with no `ITEM_ARMOR` check, when `utils.h:1714` already has the guarded form. |

**1 and 2 I would block on.** 1 is a live resource leak that the PR newly documents as handled, which is worse than leaving it undocumented. 2 turns any future assertion failure in this file into cross-test global corruption rather than a test failure.

**3, 4 and 5 are pre-existing code that moved verbatim** - I am not claiming this PR broke them. I am claiming this PR is the moment they become the published contract of a new module, and the new headers, which are otherwise unusually careful about ownership, nullability and lifetime, get all three of these preconditions wrong or silent. The whole argument for the split is that `combat_messages.h` is now the thing a caller reads instead of `fight.c`. That only pays off if the header is true.

On the decomposition itself: picking targets by measured risk instead of line count is the right call, leaving ranks 1/2/4/6 alone because their fan-in is 240-403 files is the right call, and the `define_druid_perks` and `set_fields[]` cleanups are real reductions rather than lateral moves. The seams are good. The contracts written on top of them need another pass.

src/combat/combat_messages.c
```diff
@@ -0,0 +1,703 @@
...
+               for condensed combat mode handling -zusuk */
+          act("$N parries $n's attack with $p!", ACT_CONDENSE_VALUE, ch, opponent_weapon, vict,
+              TO_NOTVICT);
+
+          /* fire any parry specs we might have */
+          spec_gateway_defense_reaction(vict, opponent_weapon, ch, "parry");
+
+          /* glance off armor */
+        }
+        else if (armor && armor_list[armor_val].armorType > ARMOR_TYPE_NONE && !rand_number(0, 2))
```
moshehbenavraham: ## 5. `armor_list[]` is indexed by a raw object value with no `ITEM_ARMOR` check

```c
struct obj_data *armor = GET_EQ(vict, WEAR_BODY);
int armor_val = -1;
if (armor)
  armor_val = GET_OBJ_VAL(armor, 1);   /* :536  "armor type" */
...
else if (armor && armor_list[armor_val].armorType > ARMOR_TYPE_NONE && !rand_number(0, 2))  /* :615 */
```

`armor_list` is `struct armor_table armor_list[NUM_SPEC_ARMOR_TYPES]` - 58 entries (`assign_wpn_armor.c:28`, `structs.h:5228`). `value[1]` only means "armor type" when the object is `ITEM_ARMOR`. For anything else that happens to carry the `ITEM_WEAR_BODY` flag it is a different field entirely: charges on a wand, contents on a drink container, dice count on a weapon.

The codebase already knows this. `src/core/utils.h:1714`:

```c
#define GET_OBJ_ARMOR_TYPE(obj) \
  ((GET_OBJ_TYPE(obj) == ITEM_ARMOR) ? armor_list[GET_OBJ_VAL(obj, 1)].armorType : ARMOR_TYPE_NONE)
```

That is the guarded version of this exact expression. This code is the unguarded one.

**Reachability, measured.** I parsed every record in `lib/world/obj/*.obj`: 1127 objects carry the BODY wear flag, 151 of them are not `ITEM_ARMOR`, and all 151 currently have `value[1]` inside `[0, 57]`. So no shipped object triggers it - it is latent, not live, and I am not claiming otherwise. It is one `oedit` session away, though: `oedit` clamps `value[1]` to `[0, NUM_SPEC_ARMOR_TYPES - 1]` only on the armor path (`src/olc/oedit.c:2744`); on any other item type the field is free-form, and negative values are not excluded either. The 151 non-armor body items also all read `armor_list[0]` today, which is a quieter correctness bug: a non-armor item is being asked a question about armor.

**Fix** - one line, using what already exists:

```c
else if (armor && GET_OBJ_ARMOR_TYPE(armor) > ARMOR_TYPE_NONE && !rand_number(0, 2))
```

and drop `armor_val` entirely, since `:536` is its only other use.

src/combat/combat_messages.c
moshehbenavraham: Fixed in 6ec215c7c. I went with the single exit.

- `trelux_claws` is tracked apart from `weap`, because the ranged and shield paths overwrite `weap`. All four `return`s and the fall-through now go through `release_claws:`, which calls `extract_obj()` (`combat_messages.c:724`). `fight.c` already uses the same `goto` pattern for `finalize_projectile`. `-Wjump-misses-init` is in the warning tier, and the build has zero warnings.
- Why I didn't stop allocating: `act()` renders `$p` through `CHECK_NULL(obj, ...)`, so a NULL object prints `ACTNULL`, and passing NULL would mean rewriting every message string that uses `weap`. Passing `&obj_proto[rnum]` would put a prototype into general `act()` code, where any future write corrupts it for the whole boot. The single exit keeps the output byte-identical.
- The header paragraph is unchanged, because it's now true.

**Test:** `Test_combat_messages_release_the_trelux_stand_in_object` gives the fixture a one-entry object table plus a zone (`read_object()` also calls `real_zone_by_thing()`). It makes the attacker a Trelux, then checks that `object_list` is NULL and `obj_index[0].number` is 0 after the call. With the `extract_obj()` removed, the test fails:

```
Test_combat_messages_release_the_trelux_stand_in_object: expected pointer <0x(nil)> but was <0x0x5b9ec00dd1f0>
```

unittests/CuTest/test_combat_messages.c
moshehbenavraham: Fixed in 6ec215c7c using the pattern you suggested: every test records its results, tears down, then asserts. I checked it with a script: none of the 17 assertions in the file comes before its teardown. That includes `CuAssertPtrNotNull(tc, singular)` in the substitution test.

The fixture now also swaps `obj_proto`, `obj_index`, `object_list`, `top_of_objt`, `zone_table`, `top_of_zone_table` and `logfile` for the new tests, and teardown restores all of them.

I reran your experiment. I flipped one assertion in `Test_combat_messages_dam_message_tells_both_participants` and added a probe after the file's tests:

```
clean run:          [PROBE] world=(nil) character_list=(nil) obj_proto=(nil) obj_index=(nil) object_list=(nil) zone_table=(nil)   OK (9 tests)
one failed assert:  [PROBE] world=(nil) character_list=(nil) obj_proto=(nil) obj_index=(nil) object_list=(nil) zone_table=(nil)   Runs: 9 Passes: 8 Fails: 1
```

The probe also printed `logfile`, which I left out above because its address changes between runs with ASLR. In both runs it pointed at the same offset inside libc's standard stream, not at the fixture's closed `tmpfile()`, so it was restored too. Neither the probe nor the injected failure is in the branch.

src/combat/combat_messages.c
moshehbenavraham: Did both, in 6ec215c7c.

1. `IS_WEAPON()` moved to `structs.h:5419`, next to `BOT_WEAPON_TYPES` and `TOP_ATTACK_TYPES`, and its `#define` is gone from `fight.c`. The two `fight.c` callers are unchanged. `dam_message()` now rejects `!IS_WEAPON(w_type)` with a SYSERR before it indexes anything (`combat_messages.c:85`), and `combat_messages.h` states the precondition.
2. In `replace_string()`, both noun-copy loops and the literal copies now stop at `buf + sizeof(buf) - 1`. While adding that bound I found the same kind of bug one line up. If a template ended in a bare `#`, `switch (*(++str))` landed on the terminator and the loop's `str++` stepped past it. A trailing `#` is now copied as plain text. No current template ends in `#`, so no output changes.

**Test:** `Test_combat_messages_dam_message_rejects_a_non_weapon_type` passes `BOT_WEAPON_TYPES`, the first out-of-range value. It checks that nothing is rendered and that the SYSERR was logged; the fixture captures `logfile`. With the guard disabled, the test process segfaults (rc 139) during rendering. So the out-of-bounds read you described can be reached through the public function, not only in theory.

src/combat/combat_messages.c
moshehbenavraham: Fixed in 6ec215c7c the "make it honest" way. `combat_messages.h` now says `dam_message()` renders nothing at `POS_DEAD`, and names where the killing-blow line and the death notice come from.

One detail changes how bad this is: the killing blow was never fully silent. After the message block, `damage_with_projectile()` has a position switch whose `case POS_DEAD` always sends "$n is dead!  R.I.P." to the room and "You are dead!  Sorry..." to the victim. The only thing that goes missing is the line describing the attack.

- `fight.c`: when `skill_message_with_projectile()` returns `SKILL_MESSAGE_MISS_FAIL` and the victim is at `POS_DEAD`, it no longer calls `dam_message()`, which is documented to do nothing there. It logs `SYSERR: damage: no death message for attack type %d` instead (`fight.c:6084`). A zero-damage hit on a living victim still falls back to `dam_message()`.
- I put the log in `fight.c` rather than inside `dam_message()`'s `POS_DEAD` branch. `dam_message()` is also reached with a dead victim from the `!dam && is_ranged` branch, which is a ranged miss against a corpse. Logging "no death message" there would be wrong.
- I didn't add a made-up death line. All 24 weapon types have message blocks in `lib/misc/messages`, so this branch can't be reached today. Deciding what new text players see for a content gap seemed like exactly the call the SYSERR is there to prompt. I can add one if you'd rather.

**Test:** `Test_combat_messages_dam_message_is_silent_for_a_dead_victim` sets the victim to `POS_DEAD` and checks that neither side is sent anything. If the melee branch is loosened to `>= POS_DEAD`, it fails. I renamed the old test to `Test_combat_messages_dam_message_tells_both_participants`, because "always produces text" was the false claim.

src/combat/combat_messages.c
moshehbenavraham: Fixed in 6ec215c7c, and `armor_val` is gone. The guarded macro at `utils.h:1714` is actually called `GET_ARMOR_TYPE_PROF()`, not `GET_OBJ_ARMOR_TYPE()`, so that is what the code uses (`combat_messages.c:638`).

A correction to the reachability numbers, because this makes it the one fix players can see. I get the same totals: 1127 objects with the BODY wear flag, 151 of them not `ITEM_ARMOR`, and none with `value[1]` outside [0, 57]. They don't all read `armor_list[0]`, though. In five of them, `value[1]` selects a real armour row:

| vnum | type | `value[1]` (max charges) | row's `armorType` | worn how |
| -- | -- | --: | -- | -- |
| 2020284-2020287 (crests and runes) | `ITEM_WAND` | 5 or 10 | LIGHT / HEAVY | `20202.zon` equips them on mobs at position 5, `WEAR_BODY` |
| 31020 (Graye's Staff) | `ITEM_STAFF` | 3 | LIGHT | TAKE + BODY, so players can wear it on the body |

Before this fix, a miss against those mobs could print "Your attack glances off a roaring lionhead crest, protecting ...". Now it prints the ordinary miss line. The change is text only:

- No spec proc is attached to any of the five objects (no `ASSIGNOBJ`), and `spec_gateway_defense_reaction()` returns early for an object without one.
- Nothing outside `combat_messages.c` reads `SKILL_MESSAGE_MISS_GLANCE`.

I've written this up in `docs/development/OVERSIZED_TRANSLATION_UNITS.md` (752636c45).

moshehbenavraham:
## Review round: all five findings addressed

Each inline thread has a reply with the fix, the reasoning, and how it was checked. Short version:

| # | Finding | Commit |
| -- | -- | -- |
| 1 | Trelux `read_object()` leak | 6ec215c7c: single exit extracts the object; test checks `object_list` and `obj_index[].number` |
| 2 | Test asserts skipped teardown via `longjmp` | 6ec215c7c: record, tear down, then assert; your probe run now shows every global restored |
| 3 | `dam_message()` unbounded `w_type` | 6ec215c7c: `IS_WEAPON()` moved to `structs.h`, guard plus SYSERR, `replace_string()` bounded; without the guard the test segfaults |
| 4 | "Always produces output" false at `POS_DEAD` | 6ec215c7c: header corrected, corpse fallback logs a SYSERR, silence pinned by a test |
| 5 | `armor_list[]` on non-armour | 6ec215c7c: `GET_ARMOR_TYPE_PROF()`; see the thread for a correction to the reachability numbers |

Two things to know before merging:

- **One fix changes what players see.** Contrary to the review's numbers, five non-armour body items do select a real armour row. Four are wand-type crests that mobs in zone 20202 wear on the body; the fifth is Graye's Staff (vnum 31020). Misses against their wearers could read "glances off" the item and now get the ordinary miss line. It is text only: no spec proc is bound to those objects, and nothing reads the glance return value.
- **I fixed a second bug next to finding 3.** While bounding `replace_string()`, I found that a template ending in a bare `#` stepped past the terminator. A trailing `#` is now copied as plain text. No current template ends in `#`.

The point in your review summary is also done. The PR body now lists the three new `LUMINARI_CUTEST` hooks among the deliberate changes. While checking that sentence I found another error in it: `act.wizard.c` gains no lines, not one.

The report records the fixes in 752636c45.

### Rebase

Not needed. The branch is 0 commits behind `master` (merge base `a71165e20`, which is master's head).

### Local CI

`scripts/ci/local/run.py` ran every job from the five workflows this PR triggers on **6ec215c7c**: **28 of 28 passed**, in 16 minutes. I didn't trust exit codes alone and checked each log for its success marker. Among them: `OK (1503 tests)` under both I/O drivers and under the sanitizers, valgrind `ERROR SUMMARY: 0 errors`, `100% tests passed` on all eight CMake builds, and `result: PASS` on both warning budgets and all five production profiles.

| Workflow | Jobs | Matrix |
| -- | --: | -- |
| `test` | 23 | `world-tools`, `build-parity`, `cmake-Debug-gcc-gcc`, `cmake-Debug-clang-clang`, `cmake-Debug-gcc-16-gcc`, `cmake-Debug-clang-22-clang-22`, `cmake-Release-gcc-gcc`, `cmake-Release-clang-clang`, `cmake-Release-gcc-16-gcc`, `cmake-Release-clang-22-clang-22`, `warning-budget-gcc-16-gcc`, `warning-budget-clang-22-clang-22`, `clean-archive-autotools`, `clean-archive-cmake`, `unit-tests`, `production-profile-autotools-gcc`, `production-profile-autotools-clang`, `production-profile-cmake-gcc`, `production-profile-cmake-clang`, `production-profile-autotools-gcc-14`, `sanitizers`, `memory-check`, `coverage` |
| `integration` | 2 | `database-migration`, `world-validation` |
| `quality` | 1 | `format-check` |
| `hygiene` | 1 | `hygiene` |
| `security` | 1 | `gitleaks` |

The last commit, **752636c45**, changes only `docs/development/`, so it triggers just `quality` and `hygiene`. I ran both on it: every format hook passed, and hygiene reports 1709 files clean. Host-side checks:

- Zero warnings under the full strict tier.
- The new tests are clean under valgrind: 0 errors, 0 bytes lost, no leaked descriptors.
- Each new test fails with its fix reverted.

### GitHub failures: all three also fail on master

- **`CMake gcc-16 Debug` / `Release`**: `vessel-memory-analyzer` and `autorun-supervision` fail inside the `gcc:16.2` container. Master's run 35028780442 at `a71165e20` fails exactly these two jobs the same way. The analyzer failure is a mawk-vs-gawk difference, since mawk rejects a ternary continued across lines in `analyze_vessel_memory_samples.sh`. The local gcc-16 jobs have gawk and pass, and this PR touches neither script.
- **`CodeQL`**: the six "new" alerts are #889-#894 (`cpp/type-confusion`, `src/core/comm.c`). They have been open on master since 2026-09-15 and are also open on #188 and #191. This PR doesn't touch `comm.c`; the check lists them because the diff is too large for it to attribute alerts, as its own summary says.

### Merging

Ready to merge. Please use **Rebase and merge**, not squash. 0e3fe4ff6 is a 13,000-line pure move that can be checked line by line, and 6ec215c7c holds the only behaviour changes. Squashing them together would make neither checkable on its own and point any bisect at the whole thing.

moshehbenavraham:
## Merged

Rebase-merged into `master`. All four commits landed separately, and the merge gave them new SHAs.
The merged tree is byte-identical to the reviewed head 752636c45: `git diff 752636c45 886703f1b`
is empty.

| On the branch | On `master` | Content |
| -- | -- | -- |
| 0e3fe4ff6 | 442020de9 | Module metrics, the ranked report, and the three pure-move extractions |
| a8d0898bb | d53383fd7 | Report links the perk defects to #194 and #195 |
| 6ec215c7c | ab49f454b | The five review fixes, and the only behaviour change |
| 752636c45 | 886703f1b | Report records the review fixes |

SHAs in the PR body and review threads refer to the branch column.

### What `master` now has

- `scripts/development/module_metrics.py`, plus `docs/development/OVERSIZED_TRANSLATION_UNITS.md`
  with the ranking, the table-heavy exceptions, and the suggested next wave.
- Three extracted modules: `src/character/perk_definitions.{c,h}`,
  `src/combat/combat_messages.{c,h}`, and `src/act/act.wizard.set.c`. Their contracts document
  ownership and lifetime, and each module has its own tests: `test_perk_definitions.c`,
  `test_combat_messages.c` and `test_wizard_set.c`.
- Fixes for the five combat messaging defects found in review:
  - A Trelux combat message no longer leaks an object.
  - `dam_message()` rejects types that are not weapons, and `replace_string()` is bounded.
  - A killing blow with no death message now logs a SYSERR.
  - The armour-glance check uses the guarded lookup.
- One change players can see: misses against wearers of the four zone 20202 crests or Graye's
  Staff (31020) no longer say the attack "glances off" the item.

### Verification

The branch was 0 commits behind `master` when it merged, so the pre-merge results cover the merged
tree:

- Local CI passed 28 of 28 jobs on 6ec215c7c.
- `quality` and `hygiene` passed on 752636c45.

The three red GitHub checks come from `master`, not this PR:

- **`CMake gcc-16 Debug` and `Release`:** both fail `autorun-supervision` and
  `vessel-memory-analyzer`. `master`'s run 35028780442 at a71165e20 fails the same two tests in
  both jobs.
- **`CodeQL`:** its six `cpp/type-confusion` annotations are all in `src/core/comm.c` (lines
  5173-5188), a file this PR does not touch.

### Issues

- #96 was already closed against this branch, and every acceptance criterion there is met. It now
  has a merge note.
- #194 and #195 stay open. Each fix changes what players see, so each needs its own review. Their
  file and test references now point at code on `master`.

### Remaining follow-up

1. **#195**, the drift in `perk_category_names[]`: 320 of 562 perks show the wrong tree name. Fix
   it before #194, because giving the barbarian perks a category runs into it.
2. **#194**, seven perk content defects. Each one has a ratchet in `test_perk_definitions.c`, so
   decrement the matching constant when you fix it.
3. **The next decomposition wave.** The report's *Suggested next wave* section lists it: an
   export audit of `utils.c`, the `has_*`/`get_*` accessors in `perks.c`, separating the command
   table in `interpreter.c` from its parser and dispatch, then three more `act.wizard.c` families.
   Leave `db.c` for last. No issue tracks this wave yet.

PR #198, opened 2026-09-16
Add paid craft trainers that advance a skill while the character is away

### LINKED ISSUE #196: Crafting: paid trainers that advance a craft or harvest skill while the character is away
## Proposal

Player proposal from Illyanna (live server), inspired by tabletop training rules: pay an expert NPC
to train one skill that has its own experience track. The character is unavailable for a fixed
real-time period (for example 24 hours) and comes back with a set amount of experience in that
skill. It need not complete a rank, especially at higher ranks, but it should do a meaningful share
of the work. Zusuk liked it as "another way to offline/AFK improve crafting at a cost".

Goals stated in the discussion:

- a gold sink for surplus gold, with fees that rise at higher levels;
- passive progress that stays "dialed back" and does not undermine character leveling;
- a reason to play other characters while one is away.

## What the code supports today

**Eligible skills.** Only craft and harvest skills have independent experience. `ability_exp[]` is
used only through `GET_CRAFT_SKILL_EXP` (`src/core/utils.h:3141`). `crafting_skill_type()`
(`src/craft/crafting_new.c:5610`) marks 12 of them active: alchemy, armorsmithing, jewelcrafting,
leatherworking, metalworking, tailoring, weaponsmithing and woodworking (craft), plus forestry,
gathering, hunting and mining (harvest). Bowmaking, cooking, fishing, poisonmaking and trapmaking
are `CRAFT_SKILL_TYPE_NONE` and are hidden from `craftscore`.

**Out of scope, following the player's own limit:**

- General skills are ranks bought at level-up and capped by character level (`GET_LEVEL + 3`,
  `src/character/study.c:3407`). Selling ranks would bypass leveling.
- Languages have no experience track. `languages_known[]` is a known/unknown flag, and
  `num_languages_learned()` (`src/character/class.c:9639`) counts every entry against the
  INT-plus-Linguistics budget, so a tutored language stored there would consume a Linguistics
  slot. That budget is also computed three different ways: the study screen
  (`src/character/study.c:2001`), the study selection handler, which alone adds Inquisitor Perfect
  Recall +3 (`src/character/study.c:3555`), and `has_unchosen_languages()`
  (`src/character/class.c:9651`), which also omits pending level-up Linguistics ranks.

**Progression math.**

- Reaching rank n+1 takes (n+1) x 1000 more experience (`craft_skill_level_exp()`,
  `src/craft/crafting_new.c:6497`): 11,000 for rank 10 to 11, and 21,000 for rank 20 to 21.
- Rank is real power. `create_craft_skill_check()` (`:3355`) refuses an item when
  20 + rank + bonuses is below its DC (10 + item level, before level adjustment), and supply orders
  pay 2 gp more per rank before bonuses (`calculate_supply_order_reward()`, `:7097`).
- Active crafting already pays well: a create takes 60 seconds and grants
  `max(50, item level x 50)` experience, limited mainly by materials. A 24-hour contract should stay
  well below an active day's gain.
- `gain_craft_exp()` (`:5135`) adds the insightful talent bonus (+5% per rank, up to +25%), awards
  one talent point per rank, and raises the rank at most once per call (a single `if` at `:5175`).
  A lump sum that crosses two thresholds gains one rank and banks the rest.

**Defect found during investigation.** Brewing bypasses that function: `src/craft/brew.c:180`,
`:205` and `:369` add alchemy experience directly. Brewing therefore never raises alchemy on its
own and never gets the insightful bonus. The banked experience only counts at a later
`gain_craft_exp()` call, one rank at a time. Training grants must share one correct entry point, so
this is a prerequisite.

**Existing mechanics to reuse.**

- **Leaving play through an NPC:** `gen_receptionist()` (`src/obj/objsave.c:1811`) saves
  belongings, sets the load room to the NPC's room, and calls `extract_char()` (`:1891`).
- **Wall-clock state without timers:** supply-order slot cooldowns persist absolute `time_t` values
  in the player file (`SuCD`, `src/player/players.c:3323`) and are evaluated lazily
  (`docs/systems/CRAFT_ACTIVITY_LIFECYCLE.md`). A training contract fits the same model: store the
  end time and settle it on the next load.
- **Marking the trainer NPC:** all 128 addressable mob flag bits are defined (`NUM_MOB_FLAGS`,
  `src/core/structs.h:1214`; 32 bits per word via `Q_FIELD`). Only index 96 (`UNUSED-96`) could be
  reclaimed, so copying the `MOB_QUARTERMASTER` pattern would spend the last free bit. A named
  registry SpecProc (`src/spec/spec_registry.c`, bound through the world `SpecProc:` field,
  `src/core/db.c:3157`) needs no flag and is already selectable in OLC. Bindings take no
  parameters, so per-trainer specialties would need separate data.
- **Placement:** after decoding all 27,129 mob prototypes in `lib/world/mob`, none carries
  `MOB_QUARTERMASTER`. The flag-gated supply-order service has no placed NPC in tracked content,
  so trainers need deliberate placement to be reachable.
- **Entry into play:** `enter_player_game()` (`src/core/interpreter.c:7693`) is called from the
  main menu (`:10001`), forced short-description setup (`src/character/char_descs.c:1015`), and
  copyover recovery (`src/core/comm.c:714`). Characters are picked in `CON_ACCOUNT_MENU`
  (`src/core/interpreter.c:8030`).
- **Policy:** `HELP MULTIPLAYING` currently allows multiplaying, so playing an alt during training
  is allowed.

**Tabletop precedent.** Pathfinder's Ultimate Campaign
[retraining](https://www.aonprd.com/Rules.aspx?ID=1469) costs 10 gp x character level x days,
takes 8 hours a day, allows one goal at a time, and rules out strenuous activity such as
adventuring meanwhile. It swaps an existing choice rather than adding progression, although
learning an additional language is one of its options. A net experience gain for gold is a
Luminari extension, which is another reason to confine it to the craft and harvest tracks.

## Proposed design

1. **Trainer NPC:** add a builder-visible "Craft Trainer" SpecProc. At a trainer,
   `apprentice <skill>` quotes the fee, duration and experience, and `apprentice <skill> confirm`
   starts the contract. The verb is a placeholder; `train`, `practice` and `study` are taken.
2. **Admission:** the skill must be one of the 12 active tracks and below the rank ceiling. The
   character must have the gold, must not be fighting, must have no primary activity running, and
   must have no pending contract.
3. **Start:** deduct the fee and store one contract in the player file (skill, experience amount,
   end time). Set the load room to the trainer's room, then save and extract as the receptionist
   does. Nothing runs while the character is away.
4. **Lock:** until the end time, the account menu shows the time remaining and the character
   cannot enter play by any path. Recalling early ends the contract and forfeits both the fee and
   the experience.
5. **Settle:** on the first load after the end time, grant the experience through
   `gain_craft_exp()`, clear the contract, and save once, so a crash cannot grant twice. The grant
   is a fixed share of the current rank's requirement, capped so one contract adds at most one rank.
6. **Fee:** the fee rises with the rank being trained, so it remains a sink at high ranks.

Left out on purpose: an in-game "busy" state, runtime timers or scans, new mob flags, trainer tiers
or specialties, languages, general skills, feats and ability scores.

## Decisions needed

- Duration (24 hours as proposed), the experience share, the fee curve, and the rank ceiling.
  Calibrate fees against real gold holdings (a read-only production query, with authorization) and
  existing sinks. Talents cost 1,000, 2,500, 5,000 or 25,000 gp base, multiplied by the next rank.
- Whether a cooldown or weekly cap is needed on top of one contract at a time. Without one, a
  crafting-only alt kept permanently in training advances every day for gold alone.
- Whether trainer experience receives the insightful talent bonus.

## Acceptance criteria

- [ ] Brewing experience goes through `gain_craft_exp()`, and a gain that crosses several
      thresholds awards every rank it reaches.
- [ ] At least one trainer is placed and reachable in world content, bound through the registry
      SpecProc.
- [ ] Admission rules and fee deduction work, and the character leaves play with belongings saved.
- [ ] A character under contract cannot enter play before the end time by any entry path, and
      recall ends the contract with no grant.
- [ ] The first load after the end time grants experience exactly once, including across a
      save and restart, and never raises the skill more than one rank.
- [ ] The contract survives a player-file round trip.
- [ ] The root CuTest suite covers the cases above.
- [ ] Help is updated in both the database and `lib/text/help/help.hlp`, and
      `docs/guides/OLC_SpecProcs.md` lists the new SpecProc.

## Possible follow-ups (out of scope)

- Language tutoring, which needs storage that does not consume Linguistics slots and one shared
  budget function.
- Trainer tiers or specialties by city.

moshehbenavraham:
Closes #196.

A player can pay a Craft Trainer NPC to advance one of the 12 active craft or harvest skills. The
character leaves play for 24 hours of real time and cannot re-enter before the contract ends. The
first time it is selected afterwards, it receives a fixed share of one rank's experience, exactly
once. Recalling early forfeits the fee and the experience. The plan, its findings, and the as-built
notes for every step are in `docs/ongoing-projects/craft-trainers.md`.

## How it works

- **Trainer.** A new world-bound `Craft Trainer` SpecProc answers `apprentice`: with no argument it
  lists each trainable skill with your rank, the experience a contract grants, and the fee;
  `apprentice <skill>` quotes and changes nothing; `apprentice <skill> confirm` takes the fee from
  gold on hand, saves the contract, and logs the character out through `perform_player_quit()`.
  It refuses untrainable skills, rank 20 and above, an open contract, combat, a running primary
  activity, too little gold, and a trainer that is asleep or cannot see the player. `ap` and `app`
  still mean `applies`; `appr` is the shortest form.
- **Contract.** Three fields in the crafting data, saved as `CrTr: <ability> <experience> <end>`
  only while a contract exists. A malformed line is logged and ignored.
- **Lock.** Main menu option 1 refuses the copy of a character that just left to train. Selecting
  the character at the account menu before the end time shows the skill, the time left, and the
  recall syntax, and returns to the menu. The account menu shows `training, 13h 20m left` or
  `training finished` in the class column.
- **Settlement.** Selecting the character after the end time grants the experience through
  `gain_craft_exp()` (so the insightful talent bonus applies) and clears the contract in the same
  save, so a crash cannot grant it twice.
- **Recall.** `recall <number>` at the account menu shows the contract; `recall <number> confirm`
  ends it with no refund and no grant.

| Rank | Grant (half the next rank) | Fee (100 x (rank + 1)^2 gold) |
| -: | -: | -: |
| 0 | 500 | 100 |
| 4 | 2,500 | 2,500 |
| 9 | 5,000 | 10,000 |
| 19 | 10,000 | 40,000 |

The tunables are macros in `src/craft/craft_training.h`. Even with the maximum insightful bonus, a
grant never crosses more than one rank; a test checks every trainable track and rank.

## Fixes this depends on

- **Brewing and multi-rank gains.** Brewing added alchemy experience directly, so it never raised
  alchemy and never got the insightful bonus. It now goes through `gain_craft_exp()`, which raises
  the skill to every rank the new experience reaches and pays one talent point per rank.
- **Respec.** `do_start()` cleared craft and harvest ranks but kept their experience and talent
  points, so each later gain paid for an old rank again; `HELP RESPEC` already promised crafting
  skills are kept. Respec now clears only general skills, and loading a character restores a rank
  its experience already earned, without paying talent points again.
- **Harvest talents.** Talent ids run to 94, but ranks were stored in 64 slots. Every mining,
  hunting, forestry, gathering, and mote synergy talent charged points and gold, wrote past the
  array, and never applied, and staff `talent set` silently did nothing for them. Rank storage now
  covers every id, and `Tlrk` writes one entry per talent; saves with 64 entries still load.
  Points and gold already spent are not refunded.
- **Copyover in the same game-loop pass as a quit.** A copyover that ran after a `quit`, `rent`,
  or new contract in the same pass (a timed copyover fires between commands and the pending
  extraction) restored the character to play and saved its already extracted belongings again,
  replacing the saved set with an empty one. The copyover writer now drops characters whose
  extraction is pending, as it drops non-playing descriptors. This replaces the plan's check in
  `copyover_recover()`, which could not have prevented the lost belongings.

## World data

`data/craft-trainers/` holds mobile 373, a master artisan with the Sanctus service NPC flags and
`SpecProc: Craft Trainer`, plus install steps for room 373 in the Sanctus III crafting district.
Like `data/harvest-tools`, it is not installed by any script. Production placement is part of the
world-data release. Install it in a development world only on a server built from this branch,
because an older server loads the unknown procedure name without a callback.

## Help and documentation

- `HELP APPRENTICE` (also `CRAFT-TRAINER`, `CRAFT-TRAINING`), a `See also` from `CRAFT-SCORE`, and
  a Craft Trainer paragraph in the builder `SPECPROC` topic, in both `lib/text/help/help.hlp` and
  new or updated `sql/components` files. Both SQL components are applied to the development
  database, and the stored entries match `help.hlp` exactly.
- `docs/guides/OLC_SpecProcs.md` lists the procedure and the current editor counts;
  `CRAFT_ACTIVITY_LIFECYCLE.md` and `SAVE_SYSTEMS_BREAKDOWN.md` cover the contract and the tags.

## Verification

- `make -j16 test` and `make install`: all gates pass, 1,526 CuTest tests. The 21 new tests are in
  `unittests/CuTest/test_craft_training.c`. Reverting any prerequisite fix, the rent save, the
  settlement call, the record clearing, the copyover condition, or the menu row makes a test fail;
  removing the entry lock makes its test crash.
- Database-backed tests (belongings saved exactly once with rent free and not free, and the
  account menu row) against an isolated MariaDB 10.11 with the CI settings: 1,526 tests, OK.
- Local CI with `scripts/ci/local/run.py --jobs 4 --cpus 4` on cec0ffa89: 28 of 29 jobs pass,
  including sanitizers, the Valgrind memory check, coverage, database migration, and clang-tidy.
  `test-production-profile-cmake-clang` failed under the matrix's load with a compiler flag probe
  error and passed when rerun alone, on this branch and on master. The two later commits change
  only comments and documentation; `quality-format-check` and `hygiene-hygiene` pass on 5f8b1c60b.
- `make test-world-tools` (542 tests), `check_build_parity.py`, and `pre-commit run --files` over
  every changed file pass. `data/craft-trainers/3.mob` validates with no findings.
- Live check on a server built from this branch, run in a private network namespace with its own
  database: listing, quoting, refusals, confirm with belongings saved, both locks, the account
  menu row, recall, and settlement, including one that crossed a rank and paid its talent point,
  and a second selection that granted nothing.

## Review notes

Review commit by commit:

1. ca4ba4dff and 1e6edf659: the plan and the owner's decisions (documentation only).
2. c2c6035c4: step 1, multi-rank gains, brewing, respec, rank restore on load, and the talent
   storage fix.
3. 3824ea360: step 2, the contract record, `CrTr`, and the pure rules.
4. 44c7e7b4c: step 3, the SpecProc and command. Mobile editor numbers past Buy Weapons shift by
   one, which accounts for the edits to the existing SpecProc registry tests.
5. 2b44280a4 and 6a69411b0: step 4, the lock, settlement, recall, account menu row, and the
   copyover writer change; then message ordering from the live check.
6. efe2a7d25: step 5, the world record.
7. 777f950e1: step 6, help and documentation.
8. cec0ffa89: clang-tidy baseline fixes (`strtol` instead of `atoi` and `sscanf`, `time_t`
   arithmetic, test fixture guards).
9. 620a7396c and 5f8b1c60b: verification record and comments.

Web players get no protocol change (decision 4 in the plan): selecting a training character
redraws the structured lobby, and the time left and `recall` appear in the classic terminal.

Please merge with **Rebase and merge** so the prerequisite fixes stay separate from the feature.


moshehbenavraham:
## Adversarial review (HEAD `5f8b1c60b`)

Traced the shipped code on every surface this review was scoped to. The PR description was not treated as proof. No gating defect was confirmed (correctness, security, races, or data loss).

Surfaces checked:

- **Contract start/save (`CrTr`)** — `craft_training_start()`, `save_char_checked()` / `load_char()` CrTr tag. Fee and record are written before quit; a failed save refunds and clears. Malformed lines are logged and ignored.
- **Gold/fee deduction** — on-hand `GET_GOLD` vs `craft_training_fee()`, `award_gold(-fee)` then save, refund on failure. Rank 0–19 fees do not overflow.
- **Entry lock out of play** — `craft_training_refuse_entry()` is the first statement of CON_MENU `1`, before intro/short-desc and `enter_player_game()`. A contract, including `training finished`, cannot enter from that menu.
- **Exactly-once settlement via `gain_craft_exp()`** — `craft_training_admit_selection()` at account-menu selection. Running contracts do not write. Finished contracts clear the record and grant in one in-memory image, then one save; a second selection grants nothing.
- **Early recall** — `craft_training_recall()`: quote does not write; confirm clears and saves with no refund and no grant. Failed save leaves the file contract intact.
- **Copyover/quit pending-extraction belongings** — `copyover_restores_descriptor()` drops `PLR_NOTDEADYET` in the same pass as quit/rent/train. The drop path does not `Crash_rentsave` again. Trainer saves belongings once (`Crash_rentsave` only when rent is not free; otherwise `perform_player_quit`).
- **Talent-rank storage past the old 64-slot bound** — `talent_ranks[MAX_TALENTS]` (128), Tlrk writes `TALENT_MAX` (95) entries inside `READ_SIZE` 512; harvest ids 67–94 and mote synergy apply and persist. Older 64-entry saves still load.
- **Multi-rank grant** — `gain_craft_exp()` raises to `craft_skill_rank_for_exp()` and pays `new_rank - old_rank` talent points. `get_craft_skill_value()` is the stored rank.
- **Brewing through `gain_craft_exp()`** — `event_brewing()` critical-fail, fail, and completion all call `gain_craft_exp(..., ABILITY_CRAFT_ALCHEMY, TRUE)`.
- **Respec rank/experience** — `init_start_char()` clears general abilities only; load restores a lagging craft/harvest rank from its experience without paying talent points again.

No inline comments: findings.md is empty.

moshehbenavraham:
Rebase-merged to `master` as `08f540e4e7` (GitHub forbids a self-approval review on this PR; the adversarial review on HEAD `5f8b1c60b` found no gating defect, and all 34 checks were green).

Still outside this PR: production placement of mobile 373, and installing `data/craft-trainers` in the main development world after the merge. An older server must not load that world record, because it would bind an unknown SpecProc name without a callback.

PR #199, opened 2026-09-16
Name and list every perk tree, and repair four perk prerequisites

### LINKED ISSUE #194: Seven perk content defects: four perks unbuyable, three invisible
## Summary

Characterisation tests added in #193 found seven content defects in the perk tables. Four make a
perk permanently unbuyable; three make a perk invisible in the perk list. None was introduced by
that PR, and none is fixed there: each correction changes what players can see or buy, which is a
content decision rather than a structural one.

The tests record them as ratchets in `unittests/CuTest/test_perk_definitions.c`, so the tables
cannot get worse while these stay open. Whoever fixes one should decrement the matching constant.

Line references are to `src/character/perk_definitions.c` on the #193 branch. Until that merges,
the same code is in `src/character/perks.c`; the `PERK_*` constants are the stable identifier
either way.

## 1. A prerequisite that no tree defines (1 perk unbuyable)

`PERK_WIZARD_EXTENDED_SPELL_3` (line 3276) requires `PERK_WIZARD_EXTENDED_SPELL_2` at rank 2:

```c
perk->prerequisite_perk = PERK_WIZARD_EXTENDED_SPELL_2;
perk->prerequisite_rank = 2; /* Must have Extended Spell II maxed */
```

`PERK_WIZARD_EXTENDED_SPELL_2` has an id in `src/core/structs.h` (`#define ... 105`) but no
definition function ever writes that slot, so it can never be acquired and Extended Spell III can
never be bought. `PERK_WIZARD_EXTENDED_SPELL_1` (line 2752) does exist, with `max_rank = 1`.

The ladder is I -> (missing II) -> III. Two plausible fixes, and the choice is a design call:

- define `PERK_WIZARD_EXTENDED_SPELL_2` so the intended three-step ladder exists, or
- repoint III at `PERK_WIZARD_EXTENDED_SPELL_1` and drop its `prerequisite_rank` to 1, since I is
  `max_rank = 1` and a `prerequisite_rank` of 2 would be unreachable there too.

Ratchet: `KNOWN_DANGLING_PREREQUISITES`.

## 2. Prerequisite ranks above the prerequisite's cap (3 perks unbuyable)

Each of these asks for rank 5 of a perk that cannot be ranked that high, so the purchase check can
never pass:

| Perk | Line | Requires | Prerequisite `max_rank` | Line |
| -- | --: | -- | --: | --: |
| `PERK_CLERIC_DOMAIN_FOCUS_3` | 4985 | `PERK_CLERIC_DOMAIN_FOCUS_2` rank 5 | 2 | 4856 |
| `PERK_CLERIC_DIVINE_SPELL_POWER_3` | 5002 | `PERK_CLERIC_DIVINE_SPELL_POWER_2` rank 5 | 3 | 4874 |
| `PERK_CLERIC_GREATER_TURNING` | 5056 | `PERK_CLERIC_TURN_UNDEAD_ENHANCEMENT_2` rank 5 | 2 | 4910 |

Fix is either lowering `prerequisite_rank` to the prerequisite's `max_rank`, or raising that
`max_rank` to 5. Which one depends on how the cleric trees are meant to be paced, so it needs a
content owner.

Ratchet: `KNOWN_UNREACHABLE_PREREQUISITE_RANKS`.

## 3. Perks with no category (3 perks invisible)

`define_barbarian_perks()` sets no `perk_category` on any of its three perks, so they keep
`PERK_CATEGORY_UNDEFINED` from the `init_perks()` reset:

- `PERK_BARBARIAN_RAGE_ENHANCEMENT` (line 8443)
- `PERK_BARBARIAN_EXTENDED_RAGE_1` (line 8458)
- `PERK_BARBARIAN_TOUGHNESS` (line 8473)

`list_perks_for_class()` skips the undefined category when printing headers, so these three never
appear under any tree in the `perk` list. They are the entire barbarian tree, and the function is a
49-line stub next to trees of 700 to 1,500 lines, so this looks like unfinished content rather than
a typo.

The categories already exist: `PERK_CATEGORY_BERSERKER` (24), `PERK_CATEGORY_TOTEM_WARRIOR` (25),
`PERK_CATEGORY_PRIMAL_CHAMPION` (26). Note that assigning any of them currently runs into a
separate display bug, #195, so fix that one first or the new headers will render incorrectly.

Ratchet: `KNOWN_UNCATEGORISED_PERKS`.

## Reproducing

```bash
make -j"$(nproc)" cutest
CUTEST_FILTER=Test_perk_definitions ./cutest
```

The tests pass today because the counts are ratcheted. Drop a ratchet constant to 0 to see the
corresponding defect fail.

### LINKED ISSUE #195: perk_category_names[] is out of sync with PERK_CATEGORY_*: 320 perks show the wrong tree name
## Summary

`perk_category_names[]` in `src/character/perks.c` has drifted out of sync with the
`PERK_CATEGORY_*` constants in `src/core/structs.h`. The array is indexed directly by the category
id, so every category from 17 upward resolves to the wrong string.

**320 of the 562 defined perks (57%) show the wrong tree name in the `perk` list.** About 250 of
them render literally as `Unknown Category`: every blackguard, inquisitor, alchemist, psionicist,
bard and paladin tree.

Found while writing up #194. Not caused by #193, which only moved the tables; the array and the
lookup both stayed in `perks.c`.

## Root cause

The enum defines 46 categories, 0 through `PERK_CATEGORY_ADAPTABLE_TACTICS` (45). The array has 36
strings plus its `"\n"` terminator. Ten categories were added to the enum without a matching
string, so everything after the first gap shifts by one and then by more:

| Missing from the array | Id |
| -- | --: |
| `PERK_CATEGORY_WAY_OF_THE_FOUR_ELEMENTS` | 17 |
| `PERK_CATEGORY_NATURES_WARRIOR`, `SEASONS_HERALD`, `NATURES_PROTECTOR` | 21-23 |
| `PERK_CATEGORY_KNIGHT_OF_THE_CHALICE`, `SACRED_DEFENDER`, `DIVINE_CHAMPION` | 27-29 |
| `PERK_CATEGORY_SPELLSINGER`, `WARCHANTER`, `SWASHBUCKLER` | 30-32 |

The stale index comments in the array show the drift starting: two entries are both commented
`/* 19 */`, and the sequence then jumps from `/* 21 */` straight to `/* 33 */`.

There is a second, independent bug in the lookup:

```c
const char *get_perk_category_name(int perk_category)
{
  /* Bounds check */
  if (perk_category < 0 || perk_category > PERK_CATEGORY_PRIMAL_CHAMPION)
    return "Unknown Category";

  return perk_category_names[perk_category];
}
```

`PERK_CATEGORY_PRIMAL_CHAMPION` is 26, which is neither the array's last real index (35) nor the
enum's maximum (45). That bound is what turns ids 27 and up into `Unknown Category`. It also means
there is no out-of-bounds read today: the too-low bound happens to keep the index inside the
array. Fixing the array without fixing the bound would still leave every category above 26 reading
`Unknown Category`; fixing the bound without fixing the array would introduce a genuine
out-of-bounds read for ids 36 to 45.

**Both must change together.**

## Impact, by category in use

| Category | Id | Perks | Displayed as |
| -- | --: | --: | -- |
| `WAY_OF_THE_FOUR_ELEMENTS` | 17 | 25 | `Hunter` |
| `HUNTER` | 18 | 16 | `Beast Master` |
| `BEAST_MASTER` | 19 | 14 | `Wilderness Warrior` |
| `WILDERNESS_WARRIOR` | 20 | 15 | `Berserker` |
| `SACRED_DEFENDER` | 28 | 15 | `Unknown Category` |
| `DIVINE_CHAMPION` | 29 | 15 | `Unknown Category` |
| `SPELLSINGER` | 30 | 16 | `Unknown Category` |
| `WARCHANTER` | 31 | 14 | `Unknown Category` |
| `SWASHBUCKLER` | 32 | 14 | `Unknown Category` |
| `MUTAGENIST` | 33 | 14 | `Unknown Category` |
| `BOMB_CRAFTSMAN` | 34 | 14 | `Unknown Category` |
| `EXTRACT_MASTER` | 35 | 14 | `Unknown Category` |
| `TELEPATHIC_CONTROL` | 36 | 14 | `Unknown Category` |
| `PSYCHOKINETIC_ARSENAL` | 37 | 14 | `Unknown Category` |
| `METACREATIVE_GENIUS` | 38 | 14 | `Unknown Category` |
| `TYRANNY_AND_FEAR` | 39 | 14 | `Unknown Category` |
| `PROFANE_MIGHT` | 40 | 14 | `Unknown Category` |
| `UNHOLY_RESILIENCE` | 41 | 16 | `Unknown Category` |
| `JUDGMENT_SPELLCASTING` | 42 | 16 | `Unknown Category` |
| `HUNTERS_ARSENAL` | 43 | 16 | `Unknown Category` |
| `INVESTIGATION_PERCEPTION` | 44 | 16 | `Unknown Category` |

Categories 1 to 16 are correct. The monk and ranger trees show a plausible but wrong name, which
is arguably worse than the obviously broken ones because nothing looks amiss.

The header is rendered by `list_perks_for_class()`, reached from `do_perk`, so this is what players
see when they type `perk`.

## Suggested fix

1. Add the ten missing strings in enum order.
2. Change the bound to the last category rather than a mid-enum constant. A `NUM_PERK_CATEGORIES`
   constant defined immediately after the last `PERK_CATEGORY_*` would stop this recurring; there
   is currently no such constant.
3. Drop the hand-written index comments, or regenerate them, since they are already misleading.
4. Add a test asserting `perk_category_names[]` has exactly one entry per category and that the
   terminator sits at `NUM_PERK_CATEGORIES`. The existing
   `unittests/CuTest/test_perk_definitions.c` is the natural home; it checks that a perk's category
   is in range but cannot currently catch a name table that is the wrong length.

## Reproducing

```bash
grep -c '"' <(sed -n '/perk_category_names\[\] = {/,/^};/p' src/character/perks.c)  # 37, incl. terminator
grep -c 'define PERK_CATEGORY_' src/core/structs.h                                  # 46
```

Related: #194 (perk content defects). Fixing the three uncategorised barbarian perks there means
assigning `PERK_CATEGORY_BERSERKER` (24), which lands in the shifted range, so this issue should be
fixed first.

moshehbenavraham:
Closes #194.
Closes #195.

The `perk` list now shows every class's perks under the right tree name, and the four perks with
broken prerequisites follow their design sheets. All eight defects in the two issues are real, but
three of them behave differently from how the issues describe them, as noted below.

## What was wrong

- **Tree names and missing trees (#195).** `perk_category_names[]` was missing ten names, so every
  tree from id 17 on showed another tree's name. `get_perk_category_name()` stopped at
  `PERK_CATEGORY_PRIMAL_CHAMPION` (26), and so did the category loop in `list_perks_for_class()`,
  which the issue does not mention. Trees 27 to 45 therefore never printed at all, rather than
  appearing as `Unknown Category`. `perk paladin`, `bard`, `alchemist`, `psionicist`, `blackguard`
  and `inquisitor` listed no perks. The monk's Way of the Four Elements showed as "Hunter", and
  the three ranger trees showed as "Beast Master", "Wilderness Warrior" and "Berserker".
- **Extended Spell III (#194).** Its prerequisite, `PERK_WIZARD_EXTENDED_SPELL_2`, was never
  defined. `can_purchase_perk()` skips a prerequisite that `get_perk_by_id()` cannot resolve, so
  Extended Spell III could be bought with no prerequisite, and Archmage of Control after it. It was
  never locked.
- **Cleric tier 3 (#194).** Domain Focus III, Divine Spell Power III and Greater Turning required
  rank 5 of prerequisites capped at 2, 3 and 2. None of them could ever be bought, and neither
  could the Divine Channeler and Master of the Undead capstones behind them.
- **Barbarian perks (#194).** The three perks had no category. They were listed at the top of
  `perk berserker` with no tree header, not hidden.

## Changes

- `NUM_PERK_CATEGORIES` follows the last `PERK_CATEGORY_*`, like `NUM_PERK_EFFECT_TYPES`. The name
  table, the lookup and the list loop all use it, and a `_Static_assert` fails the build if the
  table does not hold one name per category plus the terminator.
- The ten missing names are restored in enum order, and the stale index comments are corrected.
  The monk, druid and paladin names are the strings `perks.c` had before `4617c772a`; the bard
  names are the tree names from `BARD_PERKS.md`.
- The content choices the issues left open follow the design sheets in `docs/systems/perks/`:
  - Extended Spell II is defined as `WIZARD_PERKS.md` describes it: Controller tier 2, cost 2,
    max rank 2, +3 rounds per rank, requires Extended Spell Enhancement I at its maximum.
  - Each cleric tier 3 perk requires its tier 2 prerequisite at that perk's maximum rank (2, 3 and
    2), which matches `CLERIC_PERKS.md` and each perk's own "(max)" text.
  - The barbarian perks are back in `PERK_CATEGORY_BERSERKER`. They had that category until
    `4617c772a` replaced `define_barbarian_perks()` with a copy from before the Ravager tree.
- In `test_perk_definitions.c` the three ratchets are now strict assertions: every defined perk
  has a category below `NUM_PERK_CATEGORIES`, and every prerequisite resolves at a reachable rank.
  `Test_perk_definitions_name_every_category` checks the terminator, the lookup at both ends of the
  range, and the first name that had shifted. `Test_perk_definitions_list_every_perk_under_its_tree`
  runs `do_perk` for every class and checks that each defined perk and its tree header appear.
- `docs/development/OVERSIZED_TRANSLATION_UNITS.md` records the fixes and the corrected symptoms.

## Verification

- `make -j16 test` and `make install` on d2e77abf5: all gates pass, 1,608 CuTest tests.
- The new and strict tests fail without the fixes. Against the old `perks.c` and
  `perk_definitions.c`, four of the six perk tests fail, and the listing test reports paladin perk
  915 as the first one missing. Removing only the barbarian categories fails the header check on
  perk 750, and restoring only the cleric rank-5 prerequisites fails the prerequisite rank check.
- Local CI with `scripts/ci/local/run.py --jobs 4 --cpus 4` on d2e77abf5: 28 of 29 jobs pass,
  including sanitizers, the Valgrind memory check, coverage, database migration and the clang-tidy
  baseline. `test-cmake-Release-clang-clang` failed its production-profile check under the
  matrix's load (`migration tier dropped baseline flag -Wcast-align`, a compiler flag probe) and
  passed when rerun alone. All 29 jobs also passed on 41f7094c8, the same change before the rebase
  onto c392a4712.
- `pre-commit run --files` over every changed file passes.
- No live server check. The listing test calls `do_perk` with a real descriptor, so the output
  goes through the same protocol layer a connection uses.

## Not changed

- The Extended Spell I to III duration bonuses have no spell code behind them. That was already
  true of I and III, and II matches them. What this PR repairs is the prerequisite chain in front
  of Archmage of Control.
- `4617c772a` also dropped Extended Rage II and III, the three berserker trees (Ravager, Occult
  Slayer and Primal Warrior, 45 perks) and the 45 druid perks. Neither issue covers them, and they
  are not restored here.


moshehbenavraham (commented):
Adversarial review of d2e77abf against master (c392a471). Scope was the shipped perk list, purchase, and persistence paths this patch actually changes.

## Findings

None. No correctness, security, race, or data-loss break reproduced on this head.

## What was traced

- `NUM_PERK_CATEGORIES` (46) vs `PERK_CATEGORY_*` ids 0..45 vs `perk_category_names[]` (46 names plus the `\n` terminator) vs `get_perk_category_name()` (`>= NUM_PERK_CATEGORIES`) and `list_perks_for_class()` (`category < NUM_PERK_CATEGORIES`). Category 0 (`PERK_CATEGORY_UNDEFINED`) is a valid index ("Undefined") but the list skips an empty/undefined header; every defined perk is required to sit in 1..45. Lookup at 45 returns "Adaptable Tactics"; 46 returns "Unknown Category" and does not treat the terminator as a name.
- Extended Spell II (`PERK_WIZARD_EXTENDED_SPELL_2` = 105): Controller, cost 2, max rank 2, +3 `effect_value`, prerequisite `PERK_WIZARD_EXTENDED_SPELL_1` at that perk's live max (rank 1 of "Extended Spell Enhancement I"). `can_purchase_perk` now rejects III without II at rank 2; II without I is rejected. Duration still has no spell consumer (`get_spell_duration_bonus` does not read perks). That matches I and III and is the stated scope of this PR (the chain in front of Archmage of Control), not a newly broken gate.
- Cleric Domain Focus III / Divine Spell Power III / Greater Turning now require rank 2 / 3 / 2 of their coded tier-2 perks, which is each of those perks' `max_rank`. Divine Channeler and Master of the Undead behind them become reachable. Greater Turning's prerequisite identity remains Turn Undead Enhancement II (pre-existing `special_description`); only the unreachable rank 5 was changed.
- Barbarian Rage Enhancement, Extended Rage I, and Toughness list under `PERK_CATEGORY_BERSERKER` (24) with a Berserker header. Confirmed through `do_perk` on a berserker-level character.
- `can_purchase_perk` still skips a prerequisite `get_perk_by_id` cannot resolve. No remaining defined perk points at an empty slot, so that skip is not currently exploitable. III specifically no longer hits it.
- Save/load writes and reads `perk_id` / class / rank as integers with no re-validation. Id 105 was already allocated; filling the slot does not shift ids. A character who already holds Extended Spell III without II still has III after load and can still buy Archmage of Control; re-buying III is blocked by max rank, not by the missing II.
- `Test_perk_definitions_list_every_perk_under_its_tree` uses `strstr` on a 30-character name prefix. Same-class cross-tree substring collisions are zero, so a dropped tree still fails the test. Same-tree prefixes exist (II inside III); the wizard `do_perk` output still prints II as its own id 105.

Exercised on this head: `can_purchase_perk`, `do_perk` / `list_perks_for_class`, `get_perk_category_name`, and the perk save/load record format in `players.c`, plus the six PR CuTests (`Test_perk_definitions_*`).

moshehbenavraham:
Merged with **Rebase and merge**. The single commit landed on `master` as b6890b076; on the branch
it was d2e77abf5. The two trees are identical (`git diff d2e77abf5 b6890b076` is empty) and the
patch-ids match, so the hashes quoted in the description and in the review refer to this change.

State at merge:

- `master` had not moved since the branch was rebased onto c392a4712, so the merged tree is the
  tree that was tested.
- Every GitHub check that ran on d2e77abf5 passed, including the production-linked tests with both I/O
  drivers, ASan/UBSan, the Valgrind memory check, coverage, the clang-tidy baseline and every
  production-profile build. The local `make test` and local CI results are in the description.
- The adversarial review of d2e77abf found no problems.
- #194 and #195 were closed when this PR was opened, and both now reference b6890b076. The branch
  was deleted on merge, so the branch links in their last comments no longer resolve.

Still not fixed, as the description notes: the Extended Spell I to III duration bonuses have no
spell code behind them, and the perks that `4617c772a` dropped are not restored. Those are Extended
Rage II and III, the Ravager, Occult Slayer and Primal Warrior trees, and the 45 druid perks. No
issue covers them.

Issue #91, opened 2026-09-04
Modernization: expand sanitizer and fuzz coverage across runtime trust boundaries


moshehbenavraham:
## Why

The repository has a useful ASan/UBSan and protocol-fuzzing foundation, but its current configuration leaves compiler ambiguity and major input boundaries untested.

Audit snapshot: 2026-09-04 at `d52d2dc64fee883ae5b7d2ad000bb90b9bedc834`.

- The sanitizer workflow installs Clang but does not set `CC=clang` for the production-linked suite, so the main job uses the runner's default GCC.
- The full sanitizer suite skips the syntax/boot path.
- There is no ThreadSanitizer job despite pthread-created worker paths in the AI service and I3 client.
- Fuzzing covers one protocol parser. The CI fuzz duration is 15 seconds.
- World files, DG scripts, configuration/env parsing, JSON/service inputs, saved-data decoders, and network command boundaries do not have equivalent fuzz targets.
- Sanitizer jobs do not assert their actual compiler/runtime selection.

## Scope

- Run production-linked ASan+UBSan tests with explicit supported GCC and Clang configurations.
- Print and validate compiler identity, sanitizer flags, and linked runtimes in each job.
- Add an isolated syntax boot and short server/client smoke interaction under ASan+UBSan using an ephemeral MariaDB test service.
- Add targeted TSan coverage for AI, I3, logging, database handoff, and other shared-state paths; document narrow suppressions only after root-cause review.
- Assess MemorySanitizer feasibility separately; enable it only with fully instrumented dependencies or document why results would be unreliable.
- Extract production parsers into fuzzable APIs rather than maintaining behavior-divergent mirrors.
- Add fuzz targets for high-risk boundaries:
  - flat world and zone data;
  - DG script/token parsing;
  - configuration and dotenv parsing;
  - onboarding/I3/AI JSON and framing;
  - command/telnet input;
  - versioned binary persistence decoders.
- Maintain seed corpora and dictionaries, preserve/minimize crash artifacts, and turn every fixed crash into a regression test.
- Keep fast deterministic PR smoke runs and longer scheduled fuzz campaigns; assess suitability for OSS-Fuzz.
- Define resource/time limits and never hide sanitizer failures with broad suppressions.

## Acceptance criteria

- ASan+UBSan jobs run with both intended compilers and fail on the first unsuppressed finding.
- The syntax boot and a minimal authenticated or pre-auth client interaction run under sanitizers.
[...]

Issue #88, opened 2026-09-04
Modernization: define the platform ABI contract and add cross-architecture/libc CI


moshehbenavraham:
## Why

The codebase has an implicit Linux/x86-64/glibc ABI but no explicit portability contract or cross-build validation.

Audit snapshot: 2026-09-04 at `d52d2dc64fee883ae5b7d2ad000bb90b9bedc834`.

- All hosted CI runs on floating Ubuntu amd64 workers.
- There is no AArch64 or alternate-libc compile/test job.
- Autoconf does not canonicalize build/host/target and has no systematic endianness, type-size, large-file, or time-width policy.
- CMake has no checked-in toolchain files and uses `try_run` for crypt behavior without a cross-compilation answer.
- The supported-platform boundary in the documentation is effectively Linux/WSL2, but exact architectures, libc implementations, database client versions, and byte-order assumptions are not defined.
- Native-ABI persistence and widespread implicit conversions make silent architecture drift higher risk.

## Scope

- Publish a platform matrix covering supported and compile-only combinations: operating system, architecture, libc, compiler, MariaDB/MySQL client, endianness, pointer width, `time_t`, and required POSIX APIs.
- Add real AArch64 Linux validation, preferably both cross compilation and execution under a native or emulated runner.
- Add a musl compile job or explicitly document why musl is unsupported.
- Add Autoconf canonical host support and targeted checks for byte order, fixed-width types, large-file support, socket types, and time width.
- Add CMake toolchain files/presets for each supported cross target.
- Replace or pre-seed `try_run` checks so cross configuration is deterministic.
- Compile at least one strict job with unsigned `char` (and, where useful, a big-endian compile/emulation target) to expose signedness and serialization assumptions.
- Ensure database, crypt, threading, DNS, networking, and optional-service discovery fail with actionable configure errors.
- Define which generated data and saved formats must be portable across architectures.

## Acceptance criteria

- The support matrix is checked into the developer/build documentation.
- Every supported target has a reproducible configure/build command.
- AArch64 and alternate-libc validation are automated at the documented support level.
- Cross builds never attempt to execute target binaries during configuration.
- Endianness, integer width, file offset, and time-width assumptions are either removed, encoded in portable formats, or enforced with clear compile/configure checks.
- Unsupported platforms fail early with a useful explanation rather than obscure compile errors.

moshehbenavraham:
[...]

PR #200, opened 2026-09-16
Replace native-structure persistence with portable binary formats

### LINKED ISSUE #95: Modernization: replace ABI-dependent raw-struct persistence with versioned portable formats
## Why

Several persistence paths read and write native C structures directly. Those files inherit compiler padding, integer widths, endianness, structure layout, and partial-write behavior, which conflicts with cross-architecture modernization and makes safe schema evolution difficult.

Confirmed examples include native-structure I/O for last-entry data in `src/act.wizard.c`, board metadata/messages in `src/comms/boards.c`, house control/object elements in `src/obj/house.c`, and player conversion utilities. This is an inventory starting point, not a complete format map.

## Scope

- Inventory every project-owned `fread`/`fwrite` path and classify it as text, explicitly encoded binary, or native-ABI binary.
- Define a small portable serialization layer using fixed-width integer types and explicit byte order.
- Give each durable binary format:
  - magic bytes and format version;
  - declared byte order;
  - bounded record and string lengths;
  - optional feature/record identifiers;
  - integrity/checksum strategy appropriate to the data;
  - explicit error and recovery behavior.
- Use checked size arithmetic and bounded decoders; reject truncated, oversized, duplicate, and unknown-critical fields safely.
- Write updates atomically through a temporary file, flush/sync as appropriate, and rename only after a complete valid write.
- Preserve backward readers for existing deployed files and provide idempotent migration/backup tooling. Never reinterpret or delete legacy data silently.
- Add golden fixtures for each historical and current format and verify them on x86-64 and AArch64.
- Fuzz every new decoder and retain malformed-input regression cases.
- Document format ownership, compatibility promises, and the process for adding fields/version bumps.
- Coordinate production data migration as a separately reviewed operational step.

## Acceptance criteria

- No durable format writes a native C structure directly.
- Existing supported files load without data loss and can be upgraded with backup/rollback.
- Golden fixtures produce identical logical results on supported architectures/libcs.
- Decoders reject malformed/truncated/oversized inputs without out-of-bounds access, leaks, hangs, or partial state mutation.
- Writes are atomic and interruption-tested.
- Every format has a checked-in specification and version policy.
- New direct native-struct persistence is blocked by review/static checks.

moshehbenavraham:
## Summary

Fixes #95. The three native-structure binary files are replaced or retired:

- **Boards** (`lib/etc/board.*`) and the **house control file** (`lib/etc/hcontrol`) now use
  versioned, little-endian formats: magic, u16 version, byte-order mark, payload size, CRC-32
  (`src/core/binary_formats.c`, no game dependencies). The loaders still read the legacy x86-64
  layouts, field by field; the first save afterwards writes the new format and keeps the legacy
  bytes as `<file>.legacy-<crc32>`.
- Loads are all or nothing. A file that cannot be decoded is renamed to
  `<file>.rejected-<time>-<pid>`. Boards used to delete a corrupt file, and house boot used to save a
  pruned file over it. Saves write `<file>.tmp` and replace through `finish_file_save()` (fsync +
  rename).
- **Login log** (`lib/etc/last`): `last all` reads it through a bounded decoder. The writer (it has
  returned before writing since 2022) and its boot-time trim are removed.
- Removed rather than ported, because no data in their formats remains: `hcontrol asciiconvert`
  (a CircleMUD 3.1 binary rent reader that dereferenced NULL on today's text `.house` files) and
  `util/plrtoascii` (CircleMUD binary player files).
- `scripts/ci/check_native_struct_io.py` (in `make test` and CTest) fails on any `fread`/`fwrite` of
  elements larger than a byte, or `read`/`write` of `sizeof(struct ...)`, under `src/` and `util/`.
  On master's sources it reports exactly the 16 calls this PR removes.
- Specification, inventory of every `fread`/`fwrite` path, recovery behavior, upgrade and rollback
  runbook, and version policy: `docs/systems/BINARY_FILE_FORMATS.md`.

Found on the way and fixed, because the new tests exercise these paths:

- `board_remove_msg()` left the vacated last index entry pointing at the moved post, so
  `board_clear_all()` freed that post twice.
- `House_boot()` now skips a record whose ownership mode is outside `house_types[]`, which
  `hsedit` indexes.
- `struct house_control_rec` loses its six never-used spare fields; `struct obj_file_elem` (only
  used by the removed converter) is gone.

## Review follow-up

The adversarial review found one defect, fixed in "Reject a current binary file whose magic is
damaged":

- The decoders read any file without the magic as a legacy layout. A current `hcontrol` with a
  damaged magic and a size that is a whole number of 912-byte legacy records could decode as one
  garbage legacy record, and the save at the end of `House_boot()` then replaced the live file. A file
  with the byte-order mark at offset 6 is now current whatever its magic: the envelope check rejects
  it as `wrong magic bytes` (`BINARY_FORMAT_BAD_MAGIC`) and boot moves it aside. Boards follow the
  same rule. The new test flips every header bit of such a file, and the specification states the
  rule.

## Acceptance criteria

| Criterion | Evidence |
| -- | -- |
| No durable format writes a native C structure | Board and house encoders; last-log writer removed; the static check finds nothing in 631 files |
| Existing files load without loss; upgrade with backup/rollback | Old native reads vs new decoders on the dev copies of `board.immortal` (9 posts), `board.mortal` (2 posts), `last` (6000 records): 0 mismatches. `hcontrol` fixture written by the old native `fwrite`. Upgrade keeps one `.legacy-<crc32>` backup (tested); rollback steps in the spec |
| Golden fixtures identical across architectures | `make -C unittests/CuTest binary-formats`: 13/13 on x86-64 and on AArch64 (gcc 16.2, ELF machine 0xB7, under qemu); new CI job on `ubuntu-24.04-arm` |
| Decoders reject malformed input safely | Every truncation and every single-bit flip of the current files, every header bit flip of a current house file one legacy record long (a damaged magic is never read as legacy), malformed fields with valid checksums, counts refused before allocating; libFuzzer with ASan/UBSan (seeded, decode, re-encode, decode must match), 60 s locally with no findings; the suite under ASan/UBSan and Valgrind |
| Atomic, interruption-tested writes | Temp file, fsync, rename. A test kills a save with `SIGXFSZ` mid-write (`RLIMIT_FSIZE`): the live file is byte-identical and the partial `.tmp` is 64 bytes |
| Checked-in specification and version policy | `docs/systems/BINARY_FILE_FORMATS.md` |
| New native-struct persistence blocked | `check_native_struct_io.py` with a self-test, no baseline |

## Verification

- `make test`: 1627 CuTest tests OK plus every script gate, on the branch rebased onto master
  `b6890b076`.
- Local CI replication (`scripts/ci/local/run.py`, every job of test, integration, quality, hygiene,
  and security): 30 of 30 passed on the final head `d11ddf1a0`, including the clang-tidy gate (no
  baseline grew) and the memory check (0 Valgrind errors). The codec harness also passed 13/13 on
  AArch64 at that commit, under qemu locally and on GitHub's `ubuntu-24.04-arm` runner.
- The first matrix, and GitHub on the first push, caught a GCC 16 `-Wnull-dereference` (warning
  budget) and CodeQL alert 895 (`cpp/toctou-race-condition` on an `access()` check); "Clear the
  analyzer and GCC 16 findings in the binary format change" fixes both.
- Real-data round trip: the dev `board.immortal`, `board.mortal`, and `last` pass the fuzz target's
  decode, re-encode, decode check with ASan/UBSan. The review fix leaves the decoded status, version,
  and count of every dev `lib/etc/board.*` and `hcontrol` unchanged.

## Production migration

Not performed by this PR; it is the separately reviewed step the issue asks for. After deployment,
house boot upgrades `hcontrol` and each board upgrades on its first post or removal, logging the
legacy file it read and the backup it kept. Rollback: stop the server, move each
`<file>.legacy-<crc32>` back over `<file>`, start the previous binary. Never start an older binary
on current-format files. Details in the spec's "Upgrading and rolling back" section.


moshehbenavraham (commented):
## Adversarial review (correctness / data loss)

One surviving finding. Hunt covered the portable board/house/last-log codecs, durable load/save/upgrade/quarantine, integer/string/limit handling, leftover native I/O, and the last-log reader.

### 1. House decoder accepts a damaged current file as a valid legacy file

- **Location:** `src/core/binary_formats.c:561` (`legacy = !has_magic`); completed by `House_boot()` at `src/obj/house.c:394` and `:442`.
- **Exploit/repro:** Encode a current `LMHC` v1 file whose length is a multiple of 912 (10 houses, 39 guests, first owner 0 — the hsedit default; other houses may have real owners). Flip any of the first 32 magic bits or set `data[0]='X'`. `house_file_decode` returns `BINARY_FORMAT_OK`, version `BINARY_FORMAT_LEGACY`, count 1, garbage vnum/owner. CRC is never checked because magic failed. `House_boot` then skips the garbage record and `House_save_control()` overwrites the live file; `replace_durable_file` labels the damaged bytes a “legacy” backup. The other nine houses are gone from the live file. The PR bit-flip sweep only used the 160-byte golden, which is not 912-aligned.
- **Fix:** If the file is at least 16 bytes and the u16 at offset 6 is `0xFEFF`, treat it as a current envelope even when magic is wrong (return a current-format error). Run the legacy decoder only when the file cannot be an envelope (no BOM, and `size % 912 == 0` or size 0).

### Surfaces checked, no other surviving finding

Board magic fallback (a 1-bit flip of `LMBD` is a huge count and is refused); envelope CRC/BOM/version and exact payload size; string NUL rules; count-before-allocate; `read_durable_file` / `replace_durable_file` / `finish_file_save` / quarantine; leftover native struct `fwrite` in boards/house/last; last-log field map and torn-file rejection; house spare-field drop (intentional); documented House_boot prune of invalid records.

src/core/binary_formats.c
```diff
@@ -0,0 +1,721 @@
...
+  struct byte_reader reader = {data, size, 0};
+  enum binary_format_status status = BINARY_FORMAT_OK;
+  const unsigned char *bytes;
+  size_t stored_count, index;
+  uint32_t guest_count, guest;
+  bool legacy;
+
+  *records = NULL;
+  *count = 0;
+  legacy = !has_magic(data, size, HOUSE_FILE_MAGIC);
```
moshehbenavraham: **[bug]** `house_file_decode` treats a damaged current file as a valid legacy file.

**Location:** `src/core/binary_formats.c:561` (`legacy = !has_magic(...)`). Persistence: `House_boot()` in `src/obj/house.c:394` (decode) and `:442` (`House_save_control()`).

**Exploit / repro:**
1. Encode a current `LMHC` v1 file whose length is a multiple of 912. One such file: 10 houses, 39 guests on the first house, first-house `owner == 0` (hsedit default). Other houses may have real owners. Size is 912.
2. Flip any of the first 32 magic bits, or set `data[0] = 'X'`.
3. `house_file_decode` returns `BINARY_FORMAT_OK`, `version == BINARY_FORMAT_LEGACY`, `count == 1`, garbage vnum/owner. The payload CRC is never checked because magic failed.
4. `House_boot` treats that as a successful load, skips the garbage record, then `House_save_control()` writes an empty/wrong current file. `replace_durable_file` copies the damaged bytes to `hcontrol.legacy-<crc>` and replaces the live file. The other nine houses are gone.

Linked against `src/core/binary_formats.c`: magic `[0]='X'` on that 912-byte file decodes as one legacy record (`vnum=1128811864`, `owner=104857600000000`, `guests=0`). The same flip on a 160-byte golden is `LIMIT_EXCEEDED` / `TRUNCATED`, which is why the existing bit-flip sweep missed this.

**Fix:** If the file is at least 16 bytes and the u16 at offset 6 is `0xFEFF`, treat it as a current envelope even when magic is wrong (return a current-format error). Run the legacy decoder only when the file cannot be an envelope (no BOM, and `size % 912 == 0` or size 0). A real legacy atrium’s high 16 bits at offset 6 are 0 for any vnum `< 65536`.

moshehbenavraham: Confirmed and fixed in d11ddf1a0 ("Reject a current binary file whose magic is damaged"), on the branch rebased onto master b6890b076.

The new test reproduces this first. Your 912-byte file (10 houses, 39 guests on the first, owner 0) with a flipped magic bit decoded as `BINARY_FORMAT_OK` on the old decoder, and the test failed on that assert before the fix.

- `house_file_decode()` and `board_file_decode()` now call `is_current_file()`. A file counts as current when it starts with its magic **or** has the byte-order mark 0xFEFF at offset 6. Only a file with neither is read as a legacy layout.
- `open_envelope()` now checks the magic itself. A mismatch returns the new `BINARY_FORMAT_BAD_MAGIC` ("wrong magic bytes"), so `House_boot()` renames the file to `hcontrol.rejected-<time>-<pid>` and does not save. Boards follow the same rule.
- One correction to the fix note: LuminariMUD has vnums above 65536. The fixture atrium 103495 has a high half of 1, not 0. The rule still holds, because a legacy file matches only when its first atrium vnum is at least 0xFEFF0000 (4278124288). The largest room vnum in the dev world is 1215752100. For boards, offset 6 is the high half of a slot number, which stays below `INDEX_SIZE` (605 today).
- `Test_binary_format_damaged_magic_is_not_read_as_a_legacy_file` builds that 912-byte file and flips each of its 128 header bits in turn. Every flip is rejected with nothing allocated, and every magic flip returns `BAD_MAGIC`. So do `data[0] = 'X'` and a house file carrying a board file's magic. A board file with a zeroed magic returns `BAD_MAGIC`, where before only the count arithmetic caught it.
- The old and new decoders return identical status, version, and count on the dev copies of every `lib/etc/board.*` and `hcontrol`.
- Checks: `make test` passed 1627 tests on the rebased head. The codec harness passed 13/13 on x86-64 and on AArch64 (gcc 16.2 under qemu), and the decoder fuzzer ran 30 s with ASan/UBSan and no findings.
- `docs/systems/BINARY_FILE_FORMATS.md` ("Common envelope") states the rule and why it exists.

moshehbenavraham:
Merged into master with **Rebase and merge**. There was no squash: the five commits are separate steps (the format change, its specification, the CI findings, the baseline record, and the review fix), as in earlier multi-commit series on master. The branch was first rebased onto master `b6890b076`, and master's tree after the merge is byte-identical to the tested head `d11ddf1a0`.

Rebase and merge rewrote every hash. Hashes quoted in the review thread and the PR body map like this:

| Branch | Master | Commit |
| -- | -- | -- |
| `bfbea16dc` | `ce4fe1681` | Replace native-structure persistence with portable binary formats |
| `5cde4f853` | `5e7fc7af4` | Document the durable binary file formats |
| `1cc37f776` | `d0d6d2635` | Clear the analyzer and GCC 16 findings in the binary format change |
| `33084fc9c` | `e36d9d628` | Record the clang-tidy findings removed with the native-structure code |
| `d11ddf1a0` | `638ba31a8` | Reject a current binary file whose magic is damaged |

**Review.** The one finding, a current `hcontrol` with a damaged magic being read as a legacy file, is fixed in `638ba31a8`. The reply in the thread gives the details, and the thread is resolved. A file with the byte-order mark at offset 6 is now current whatever its magic. Such a file is rejected as `BINARY_FORMAT_BAD_MAGIC` and moved aside at boot, and boards follow the same rule. The new test reproduced the defect on the old decoder before the fix.

**Verification on `d11ddf1a0`:**

- Local CI replication (`scripts/ci/local/run.py`): 30 of 30 jobs passed. Each log shows its success marker: 1627 tests in the unit, sanitizer, memory-check, and coverage jobs, "clang-tidy baseline respected", "budget respected" for both warning budgets, 34/34 CTest runs, and both fuzzers completing.
- GitHub: 61 checks passed. `Dependency Review` was skipped by its `if:` condition. `Binary file formats (AArch64)` ran 13 tests OK on `ubuntu-24.04-arm`.
- `make test`: 1627 tests OK. The codec harness passed 13/13 on x86-64 and on AArch64 (gcc 16.2 under qemu).

One matrix run before the final push was on an unpublished version of the review fix, and it failed two jobs:

- The clang-tidy gate flagged `bugprone-not-null-terminated-result` on a `memcpy` of the magic string literal in the new test. The test now copies the magic from the board fixture, and the fix was amended in before the push.
- `test-cmake-Debug-clang-clang` failed CTest's `production-profile` test with "migration tier dropped baseline flag -Wcast-align". That is a known flake of the warning probe under heavy container load. The job passed alone, and again in the final matrix.

**Still open:**

- The production migration of `lib/etc/board.*` and `lib/etc/hcontrol` is not part of this PR. It is the separately reviewed operational step #95 asks for, and `docs/systems/BINARY_FILE_FORMATS.md` ("Upgrading and rolling back") has the steps.
- #88 and #91 get comments on the parts of this PR they can build on: the AArch64 job and portable formats for #88, the persistence-decoder fuzz target for #91.

PR #190, opened 2026-09-15
docs: document current crafting systems

dubstylee:
## Summary

- replace the partial crafting notes with a source-backed reference for runtime routing, legacy catalog and kit systems, materials-and-motes projects, timed activities, persistence, and known limitations
- synchronize 17 player-help topics and 40 managed aliases between `lib/text/help/help.hlp` and the SQL migration/verifier
- keep SQL updates idempotent and tag-scoped so unrelated help owners survive repeated application
- align lifecycle, event inventory, OEDIT, generated web-guide, and historical wilderness documentation with current runtime source

This remains documentation/help-only. It does not change crafting gameplay, repair runtime accounting behavior, enable dormant handlers, add world objects, or select a production crafting mode.

## Third review round

The third cleanup pass addresses all 16 new inline findings and preserves four earlier current-head regression targets. Notable corrections include:

- warn that general `supplyorder reset|abandon|cancel` can discard shared project allocations without refund, while `select` replaces unstarted project metadata without itself clearing old allocations
- remove unsupported public resource-accounting guarantees without documenting the privately reported reproduction
- distinguish category-harvest quality tools, equipment admission slots, unreachable node-harvest rechecks, and workflow-specific station checks
- document standalone versus kit `reforge`, including interception by carried, worn, or room kits
- warn that unstoring and restoring higher-grade hides downgrades them to low-grade hide
- correct material/bonus/mote ordering and repeated material-group allocation in crafting examples
- document that `supplyorder list` shows points rather than generated offers and that selection/cooldown perform lazy refresh
- distinguish stored crafting values `0|1|2` from `cedit` menu choices `1|2|3`
- document reset-selector abbreviations and cross-field side effects
- correct material-golem balance checking, ash-wood completion behavior, and registered golem animation syntax
- qualify stored-skill, rapid-talent, station, and efficient-savings chronology, including state mutation by `show`
- document the builder-bindable third supply-order handler without claiming a deployed binding
- identify the mold-vendor special as unassigned/unregistered while preserving deployment boundaries
- remove the player-inaccessible `REDESC` help link
- correct OEDIT station/refining descriptions and regenerate the Pandoc mirror

## Fourth review round

The final six inline comments are also addressed:

- correct the synthetic craft lifecycle evidence: committed scripted relocation is covered, but entry-script rollback with an attached craft is not
- remove the unsupported claim that craft tests prove material reservation behavior
- document that a natural-1 critical failure performs the whole-project reset without reimbursement, rather than clearing only allocations
[...]

Issue #92, opened 2026-09-04
Modernization: raise risk-weighted test coverage and eliminate randomized CI variance


moshehbenavraham:
## Why

The production-linked CuTest suite, protocol harness, Valgrind job, coverage job, and syntax boot are strong foundations. The current coverage contract is nevertheless too low and too variable to support a large warning/API/serialization modernization safely.

Audit snapshot: 2026-09-04 at `d52d2dc64fee883ae5b7d2ad000bb90b9bedc834`.

- The CI floor is 10.50% line coverage and 7.16% branch coverage.
- Coverage documentation notes that randomized execution can make measured counts vary.
- A single repository-wide floor does not protect changed code or critical authentication, parsing, persistence, database, networking, OLC, and concurrency paths.
- Release-optimized behavior is not validated as a distinct test profile.

## Scope

- Seed every random source used by tests and print the seed; make the default CI seed deterministic.
- Add an explicit repeat/stress mode that exercises multiple recorded seeds without changing the blocking coverage baseline nondeterministically.
- Add changed-line coverage and subsystem-specific thresholds for security- and data-critical code.
- Establish a monotonic ratchet from the current measured baseline instead of selecting an arbitrary aspirational percentage.
- Add focused tests for:
  - authentication and account lifecycle;
  - command and numeric parsing boundaries;
  - SQL bind/result/error paths;
  - world/DG/config parsers;
  - persistence compatibility and malformed inputs;
  - networking and descriptor lifecycle;
  - OLC read/write round trips;
  - threaded service shutdown and handoff.
- Run the production-linked suite in an optimized/Release configuration in addition to the debug/coverage build.
- Extend the live-server smoke test beyond listening-port detection to a minimal deterministic protocol interaction with isolated database state.
- Publish coverage artifacts with path normalization and exclude only generated/vendor code through reviewed rules.
- Add mutation testing or targeted fault injection for a small set of critical parser/auth modules to measure assertion quality.

## Acceptance criteria

- Re-running the same coverage job at the same commit produces the same seed and stable totals.
- New or changed critical code cannot lower its subsystem or changed-line coverage threshold.
- Repository line and branch floors only increase.
- Debug, optimized, sanitizer, and coverage profiles all execute the production-linked suite.
- Critical subsystem gaps have named tests and owners, not only a global percentage.
- CI retains human-readable and machine-readable coverage artifacts.
- A documented stress job can replay any failing random seed.

[...]

PR #201, opened 2026-09-16
Fix crafting projects, supply orders and reforge in the materials-and-motes system

moshehbenavraham:
Follow-up to the crafting code review done during #190. This fixes the defects that review found in
the materials-and-motes crafting system (`crafting_system 2`), plus defects with the same causes
that turned up while tracing them. The commits are grouped by area. Eight address the review on
this pull request, and the last one aligns the crafting help and notes that #190 merged with these
fixes.

## Boot data

`crafting_recipes[]` and `materials_sort_info[]` were only filled by a boot block for the retired
campaign variants. With `crafting_system 2`, no project could choose a variant, supply orders could
not pick recipes, and `craftmaterials` listed no materials. `boot_db()` now fills both.
`sort_materials()` no longer sorts one element past its array, and the listing starts at the first
material instead of skipping it. The world boot test now checks the recipe table.

## Craft projects

- `craft` subcommands that change the project wait while craft work runs (`show`, `check` and
  `score` still work), and completion makes an item only if the project is still fully paid.
- `craft show` and failed completions extract the objects they build. The four per-type completion
  routines are now one.
- Efficient-talent savings end with the materials they came from.
- Removing a bonus slot's motes refunds only that slot, and the message shows the amount returned.
  `craft motes` accepts only bonus slots 1 to 6.
- The recipe variant's skill is the project skill for the tool, station, talents, roll and
  experience, as supply orders already used it, so a first attempt no longer depends on whether
  `craft show` ran. `craft check` reports a missing station, and `craft start` refuses a DC no roll
  can reach.
- Bonuses work on a recipe's first variant, `itemtype` offers only what recipes can build, and
  `specifictype` lists a weapon or armor type only when a recipe variant builds it. The hand
  crossbow recipe is filed under its own type instead of the heavy crossbow's. Room rnum 0 counts
  as a room, the resize reset clears the size, the busy message can name golem work, and the
  natural 1 message says the project is lost.
- Pickups under `USE_NEW_CRAFTING_SYSTEM` store by crafting material instead of object material,
  with the same lookup as `craftmaterials store`. That option is off in the default build; the
  pickup path was compiled and run once with it.

## Supply orders

- Craft project commands and supply order commands no longer act on each other's project; the two
  share one project record. While an order is held, `craft score` and the golem commands still
  work, but golem construction cannot start.
- Abandoning or finishing an order returns the materials no finished item used, every new order
  starts with no progress, and completion needs exactly the ordered quantity.
- An item takes its materials only when every group can pay for it.
- Orders and offers pick weapon, armor and misc recipes only, as the code comments intended, and an
  offer without a usable variant cannot be selected; the next refresh replaces it.
- A selected contract keeps its type and quality tier across saves (`CrCT`), and a quality contract
  takes only materials graded above its tier.
- The reward's skill bonus reads the recipe's crafting skill, material subcommands stop after a
  refusal, the New Supply Orders special runs the `supplyorder` command itself, and player text no
  longer points at a `supplyorder list` that shows no offers.

## Golems, storage and reforge

- A golem project is saved (`CrGo`), so construction resumed after login or copyover can finish, and
  a wooden golem keeps the wood type found when construction starts. Golem work that finishes or is
  cancelled clears its work method, so it no longer blocks supply orders until `craft reset`.
- Bundles from `craftmaterials unstore` record their crafting material, so hide grades survive
  storing them again.
- Both reforge paths pass the player's text to `is_abbrev()` first, so abbreviations and the
  numbered composite bows match, and the lookups skip the undefined type 0. Kit reforge, restring,
  redesc and resize count every item in the kit.

## Decisions for review

- Woodworking has no tool slot, so carpentry needs no worn tool. The carpentry table still applies.
- Following their recipes, bows, staves and instruments now roll woodworking, and leather or cloth
  pieces roll tailoring, where weapons and armor always rolled weaponsmithing or armorsmithing.
- A natural 1 still loses the whole project, as a success consumes it; only the message changed.
- A quality contract requires materials graded above its tier: superior needs grade 2 or better,
  exceptional grade 3 or better.
- `supplyorder select` keeps working by number. The offer listing removed in e90097d92 is not
  restored.
- The blowgun and the duplicate chainmail helm type have no recipe, so `specifictype` no longer
  offers them; no recipe was added.

## Help and docs

#190 merged help and notes that describe the crafting code before these fixes, so the last commit
updates them. Thirteen topics change in `lib/text/help/help.hlp` (CRAFTING, CRAFT-ITEMTYPE,
CRAFT-MATERIALS, CRAFTING-RECIPES-MATERIALS, VESSELS-AND-ADVANCED-CRAFTING, NEWCRAFT, CRAFT-SHOW,
CRAFT-CHECK, CRAFT-BONUSES, CRAFT-SPECIFIC-TYPE, CRAFT-VARIANT, SUPPLYORDER and REFORGE) and are
regenerated in `sql/components/help_crafting_entries.sql`, with new body hashes in its verifier. The
`help_crafting_project_entries.sql` component this branch first added is gone: its text
replacements targeted the topics as they were before #190.
`docs/world_game-data/CRAFTING_SYSTEM_NOTES.md`, `docs/world_game-data/OEDIT_GUIDE.md` (with the
generated `oedit.html`) and `docs/systems/CRAFT_ACTIVITY_LIFECYCLE.md` describe the fixed behavior,
and `docs/systems/SAVE_SYSTEMS_BREAKDOWN.md` lists the `CrCT` and `CrGo` tags.

## Verification

- Rebased onto master at a8a6ab71a (#190). The conflicts were in `Makefile.am`,
  `sql/components/ci_schema_manifest.txt`, `lib/text/help/help.hlp` and
  `docs/systems/CRAFT_ACTIVITY_LIFECYCLE.md`. `src/`, `unittests/` and `CMakeLists.txt` are
  identical to the reviewed head b5e6224bd.
- `make -j$(nproc) test` on the host passed at the rebased head: 1657 CuTest tests plus the shell
  gates.
- `help_crafting_entries.sql` was applied twice to temporary copies of the development help tables,
  and its verifier passed all 37 checks.
- New tests are in `unittests/CuTest/test_crafting_projects.c` (28) and `test_craft_training.c`
  (two save and load tests), and the world boot test checks the recipe table. The tests for the edit
  lock, the completion re-check, recipe picks, fresh order progress and the boot load, and the test
  for each review fix, were run against the code without their fix and failed. The pickup fix has no
  committed test because the option is off; a scratch build with it failed the same way before the
  fix and passed after.
- Every CI job this pull request triggers was run locally at b5e6224bd with
  `scripts/ci/local/run.py`: all 30 passed, including `quality-clang-tidy` ("clang-tidy baseline
  respected"). At the rebased head eb1576b70, the ten jobs whose inputs the rebase changed passed
  again: format check, hygiene, gitleaks, database migration, world validation, build parity, both
  clean-archive builds, world-data tools and the production-linked tests. GitHub-only CodeQL was not
  run locally.
- No live server run: the tests call the real command handlers.

## Not changed

- `material_grade()` has no grade for brass, cotton, bone and dragonblood.
- The supply reward's material bonus reads allocations after the last item has used them, so it is
  0 at completion.


src/craft/crafting_new.c
```diff
@@ -1481,7 +1483,7 @@ void set_crafting_motes(struct char_data *ch, const char *argument)
     }
     send_to_char(
         ch, "You recover %d %ss from your project. Type craft show to review your projects.\r\n",
-        required, crafting_motes[mote_type]);
+        allocated, crafting_motes[mote_type]);
```
moshehbenavraham, line 1489: **Pre-existing, but in the remove path this PR edits: `craft motes add|remove 7` reads and writes past the six bonus slots, and remove mints motes.**

The slot check at line 1349 is `slot > (MAX_OBJ_AFFECT + 1)`, so slot 7 passes and becomes index 6 after `slot--`. In `struct crafting_data_info`, `affected[6]` reads `motes_required[0..3]`, and `motes_required[6]` is the `enhancement` field right after the array.

Reproduced on this branch (long sword project, bonus slots 1 and 2 each holding 3 motes, so `affected[6]` reads as location 3 = `APPLY_INT` with modifier 3):

```
craft enhancement 8   -> You set your project's enhancement bonus to 8.
craft motes remove 7  -> You recover 8 ice motes from your project.
```

Repeating the pair took ice motes 0 -> 8 -> 16 -> 24, and each remove reset the enhancement to 0 so it could be set again for free. The add side corrupts the project too: with the enhancement at 0, `craft motes add 7` took 3 ice motes and set `enhancement` to 3.

Fix: `if (slot < 1 || slot > MAX_OBJ_AFFECT)`, matching `set_crafting_bonuses()`.

src/craft/crafting_new.c
```diff
@@ -4586,6 +4442,13 @@ void newcraft_create(struct char_data *ch, const char *argument)
     send_to_char(ch, "%s", NEWCRAFT_CREATE_NOARG1);
     return;
   }
+  /* A held supply order occupies the project record (its item type is an object type). */
+  else if (player_has_supply_order(ch))
```
moshehbenavraham: **This guard also refuses `craft score` and every `craft golem` subcommand while an order is held.**

It runs before the whole dispatch, so while an order is held, `craft score`, `craft golem show|type|size|reset` and `craft golem animate <corpse>` all print "You are working on a supply order..." (confirmed on this branch). None of them touch the shared record: `show_craft_score()` only reads ranks, the golem setters write only `golem_*`, and `animate_bone_golem()` never touches `GET_CRAFT()` (its bone and motes are the separate owned counts).

Of the golem subcommands, only `craft golem start` needs the refusal, because `start_craft_activity()` overwrites `crafting_method` and would drop the order. Suggest letting `score` and `golem` through, with the check inside `begin_golem_craft()` or the `start` branch of `newcraft_golem()`.

moshehbenavraham (commented):
Reviewed at b3118593f.

Traced and found sound:
- The edit lock is needed: `craft` declares STANDARD|MOVE, which don't conflict with craft work's HANDS|ATTENTION.
- `CrGo` resume works because `CrDC` and `CrSk` are already saved. Bone golems can't be built by command, so the `<= GOLEM_TYPE_IRON` bound is right.
- No material object in `lib/world` sets value 1, so the new `craftmaterials store` lookup can't misread world items.
- Simulating `select_stable_craft_recipe()` for every reachable seed never produced an instrument offer.
- The five help entries in the development database match `help.hlp`.
- CI is green, including the clang-tidy baseline.

Each finding below was reproduced with a scratch program linked against this branch's cutest objects (nothing committed), except the `USE_NEW_CRAFTING_SYSTEM` one, which was found by reading.

- New in this PR: the supply order guard also blocks `craft score` and `craft golem`; refused stale offers never refresh; the pickup path still loses hide grades.
- Reachable because of this PR: the hand crossbow and blowgun recipes.
- Pre-existing, in edited code or next to the PR's goals: bonus slot 7 mints motes (the most serious, and a one-line fix in a function this PR already edits); a finished golem blocks supply orders; the New Supply Orders special swallows subcommands.
- One comment-style nit.

src/craft/crafting_new.c
```diff
@@ -9103,6 +9020,14 @@ int select_contract_by_id(struct char_data *ch, int contract_id)
 
   struct supply_contract *contract = &contracts[contract_id - 1];
 
+  /* Saved offers from older versions can name an instrument recipe without a variant. */
```
moshehbenavraham: **A refused stale offer stays active, so that slot never refreshes.**

This check blocks the bad selection, but it leaves the offer active. `refresh_supply_slots()` only fills inactive slots, and `reject_contract_by_id()` has no caller. A save holding an old no-variant instrument offer (the old `% 4` type pick could produce one) keeps it for good: selecting that number always answers "That contract is not available.", and only the other four slots ever refresh.

Reproduced on this branch: slot 0 held the lyre with variant -1 and a refresh was due. Two `supplyorder select 1` calls were both refused, and after each refresh slot 0 was still active with recipe 138 and variant -1.

Suggest deactivating the matching slot here, or dropping active slots with an invalid recipe or variant at the top of `refresh_supply_slots()`, so the next hourly refresh replaces it.

src/craft/crafting_new.c
```diff
@@ -869,7 +871,7 @@ static void set_craft_weapon_type(struct char_data *ch, char *arg2)
 
   for (i = 1; i < NUM_WEAPON_TYPES; i++)
   {
-    if (is_abbrev(arg2, weapon_list[i].name))
+    if (is_valid_craft_weapon(i) && is_abbrev(arg2, weapon_list[i].name))
```
moshehbenavraham: **Two weapon types still have no reachable recipe, and one recipe builds the wrong weapon.**

- `blowgun` (80) has no recipe.
- `hand crossbow` (60): `crafting_recipes.c:455-456` passes `WEAPON_TYPE_HEAVY_CROSSBOW` as the practical type, so `set_crafting_variant()` files it under heavy crossbow.

Reproduced on this branch:
- `craft specifictype blowgun` and `craft specifictype hand crossbow` are accepted, but `craft variant` then lists nothing, so the player can only `craft reset`.
- `craft specifictype heavy crossbow` lists both "heavy crossbow" (6 wood) and "hand crossbow" (3 wood). Choosing "hand crossbow" sets recipe 61 while `crafting_specific` stays 12, so `setup_craft_weapon()` builds a heavy crossbow at half the wood.

The data is old, but it only became reachable when this PR started filling `crafting_recipes[]` at boot. Suggest `WEAPON_TYPE_HAND_CROSSBOW` as the practical type at `crafting_recipes.c:456`, and excluding blowgun in `is_valid_craft_weapon()` (or adding a recipe).

Minor: armor type 26 (`SPEC_ARMOR_TYPE_CHAIN_HEAD`, a duplicate "chainmail helm") is also listed with no recipe. Typing the name always selects type 25, so it only shows twice in the list.

src/craft/crafting_new.c
```diff
@@ -8136,13 +8063,8 @@ void request_new_supply_order(struct char_data *ch)
     return;
   }
 
-  if (GET_CRAFT(ch).crafting_item_type > CRAFT_TYPE_NONE || GET_CRAFT(ch).crafting_specific > 0 ||
-      GET_CRAFT(ch).craft_variant >= 0)
-  {
-    send_to_char(ch, "You already have a supply order going. Type supplyorder show to see the "
-                     "details or supplyorder reset to start over.\r\n");
+  if (craft_project_blocks_supply_order(ch))
```
moshehbenavraham, line 8087: **Pre-existing, next to this PR's separation of the two record users: a finished or cancelled golem blocks supply orders until `craft reset`.**

`craft_golem_complete()` finishes the project through `reset_current_golem_craft()`, which clears only `golem_*`, and `craft_activity_ended()` clears only `craft_duration`. So `crafting_method` stays `SCMD_NEWCRAFT_GOLEM` after the golem is built or the work is cancelled. The unchanged check at line 8060 then answers `supplyorder request` with "You are already working on a crafting project." The same happens for `select` at line 7443. `supplyorder abandon` reports no order, so the only way out is `craft reset`, which also discards an idle item project.

Reproduced on this branch: after the golem reset, method 8 and duration 0 gave exactly those two messages.

Suggest clearing `crafting_method` when a golem project ends, for example at the end of `craft_golem_complete()` or in `reset_current_golem_craft()` when the method is `SCMD_NEWCRAFT_GOLEM`.

src/obj/act.item.c
```diff
@@ -2294,9 +2294,10 @@ static void get_check_craft_material(struct char_data *ch, struct obj_data *obj)
     return;
 
   quantity = MAX(1, GET_OBJ_VAL(obj, 0));
-  material_type = GET_OBJ_MATERIAL(obj);
+  /* Storage is indexed by crafting material, not by object material. */
+  material_type = obj_material_to_craft_material(GET_OBJ_MATERIAL(obj));
```
moshehbenavraham: **This pickup path still loses hide grades.**

This converts only the object material, so it misses the value 1 lookup that `craftmaterials store` got in this PR. With the option on, a high grade hide bundle from `craftmaterials unstore` that is dropped and picked up again has `MATERIAL_LEATHER` and is stored as `CRAFT_MAT_LOW_GRADE_HIDE`.

Suggest sharing the lookup in `do_list_craft_materials()`: use value 1 when it maps back to the object's material, otherwise convert. The option is off in the default build, so I found this by reading only.

src/craft/crafting_new.c
```diff
@@ -8448,12 +8368,9 @@ SPECIAL(new_supply_orders)
   {
     complete_supply_order(ch);
   }
-  else if (is_abbrev(arg1, "reset"))
-  {
-    reset_supply_order(ch);
-  }
-  else if (is_abbrev(arg1, "abandon"))
+  else if (is_abbrev(arg1, "reset") || is_abbrev(arg1, "abandon"))
```
moshehbenavraham: **Latent and pre-existing: this special swallows every `supplyorder` subcommand it doesn't handle.**

No mob in this checkout's `lib/world` uses "New Supply Orders", but builders can attach it. It returns 1 for every `supplyorder`. In that mob's room, `select`, `list`, `cooldown`, `materials` (only `material` matches), `status`, `begin`, `finish` and `cancel` print the usage text instead of reaching `newcraft_supplyorder()`.

Now that this PR sends `reset` and `abandon` here to the shared helper, delegating the whole command to `newcraft_supplyorder()` would stop the two entry points from drifting.

src/craft/crafting_new.c
```diff
@@ -3360,38 +3361,20 @@ bool create_craft_skill_check(struct char_data *ch, struct obj_data *obj, int sk
...
+  skill_mod = get_craft_check_modifier(ch, skill);
 
   if ((20 + skill_mod) < dc)
   {
     send_to_char(ch, "You don't have the skill to craft %s.\r\n", obj->short_description);
     return FALSE;
   }
 
-  // critical failure. Lose the item, materials and motes.
+  // critical failure. The whole project is lost, as a success would consume it.
```
moshehbenavraham: Nit (AGENTS.md: keep `/* */` comments): this edited comment and the ones at lines 3985, 7923 and 10084 are added lines that still use `//`.

moshehbenavraham:
Cross-link: #92 (deterministic test seeding and coverage ratchets) now has branch [`test/92-risk-weighted-coverage`](https://github.com/LuminariMUD/Luminari-Source/tree/test/92-risk-weighted-coverage).

`Test_supply_order_requests_start_fresh_with_orderable_recipes` in this PR calls `circle_srandom(1)` and then reseeds from `time(NULL)`, the same pattern as the 24 clock reseeds in 8 test files on `master`. The first scope item of #92 is to seed every random source used by tests, print the seed, and make the default CI seed deterministic. Nothing needs to change here now: whichever of this PR and the #92 branch merges second moves this test onto the new seeding rule.

src/craft/crafting_new.c
moshehbenavraham: Fixed in 34e7f7b22 (Refuse bonus slot 7 in craft motes): the check is now `slot < 1 || slot > MAX_OBJ_AFFECT`, as in `set_crafting_bonuses()`.

`Test_craft_motes_refuse_a_bonus_slot_past_the_sixth` sets up your layout (bonus slots 1 and 2 holding 3 motes each, enhancement 8, 100 of every mote), then checks that `craft motes remove 7` and `craft motes add 7` are both refused and leave every mote count, the enhancement and both slots unchanged. It fails with the old bound.

src/craft/crafting_new.c
moshehbenavraham: Fixed in 6a1b4a55a (Let craft score and golem commands run beside a supply order). The guard before the dispatch is gone. A held order is now refused only where the project record is read or changed:

- `craft_project_is_busy()`, which every project-changing subcommand already calls, refuses a held order first;
- `show` and `check` refuse it;
- `begin_golem_craft()` refuses it at the top, since `start_craft_activity()` would record the golem method over the order's.

`craft score` and `craft golem type|size|show|reset|animate` run normally. The dispatch order is unchanged, so abbreviations such as `craft s` still reach `specifictype`.

`Test_craft_commands_leave_a_supply_order_alone` now also checks `check`, `score` and `golem type wood`, and that `golem start` with the feat, wood, bronze and motes on hand is refused, starts no activity and keeps the order. It fails on the old guard, and on the new code without the `begin_golem_craft()` refusal. `docs/systems/CRAFT_ACTIVITY_LIFECYCLE.md` states the rule.

src/craft/crafting_new.c
moshehbenavraham: Fixed in ee5d9df2d (Replace a supply offer that has no recipe variant). `refresh_supply_slots()` first deactivates any active offer without a usable recipe variant, so the refresh fills that slot like any empty one. The check is shared with `select_contract_by_id()` as `supply_contract_is_orderable()`; the select refusal stays for a stale offer seen before a refresh is due.

`Test_supply_order_offers_always_have_a_variant` now makes the refresh due after the refused select and checks that `supplyorder select 1` then takes a new offer that has a variant and is not the lyre. It fails without the change.

`Test_gameplay_supply_refresh_is_lazy_and_preserves_existing_offers` built its five preserved offers with recipe 0, which the refresh now replaces, so those offers name a recipe and variant. I left `reject_contract_by_id()` alone, since nothing calls it.

src/craft/crafting_new.c
moshehbenavraham: Fixed in 2e6df8313 (Offer only the weapon and armor types a recipe builds).

- The hand crossbow recipe's practical type is now `WEAPON_TYPE_HAND_CROSSBOW`.
- Instead of adding the blowgun and armor type 26 to hard-coded lists, `is_valid_craft_weapon()` is replaced by `craft_specific_type_has_recipe()`: `craft specifictype` lists and accepts a weapon or armor type only when a recipe variant is filed under it. That hides the blowgun and type 26 and still hides the same composite bows. No blowgun recipe was added.

A scratch enumeration of every weapon, armor, misc and instrument type against the loaded recipe table found exactly the gaps you listed, and none for misc items or instruments, so those two lists are unchanged.

`Test_craft_specific_types_offer_the_recipes_that_build_them` checks that every weapon and armor recipe is filed under the type it builds, that `specifictype blowgun` is refused, that the heavy crossbow's `craft variant` list no longer offers the hand crossbow, that `variant hand crossbow` sets recipe 61 on type 60, and that the armor list shows `chainmail helm` only once apart from the light one. The old recipe data, the old weapon filter and an unfiltered armor list each fail it.

src/craft/crafting_new.c
moshehbenavraham: Fixed in f2b89034a (Clear a golem project's method when its work ends).

- `reset_current_golem_craft()`, which every completion path calls, clears `crafting_method` when it is `SCMD_NEWCRAFT_GOLEM`. That also lets `craft golem reset` clear a stale method saved before this fix.
- `craft_activity_ended()` does the same for cancelled golem work. The cancelled project keeps its golem type and size for another start.

`Test_golem_work_that_ends_leaves_supply_orders_available` cancels started golem work, then finishes a golem project through `craft_golem_complete()`, and requests a supply order after each. Removing either half of the fix fails its path. `docs/systems/CRAFT_ACTIVITY_LIFECYCLE.md` describes it.

src/obj/act.item.c
moshehbenavraham: Fixed in 9555e1ae9 (Keep a bundle's hide grade when it is picked up). The lookup is now `craft_material_from_object()` in `crafting_new.c`, and both `craftmaterials store` and `get_check_craft_material()` call it.

The option is off in the default build, so there is no committed test. I linked a scratch cutest with `act.item.c` compiled with `-DUSE_NEW_CRAFTING_SYSTEM` and ran a test that unstores 5 high grade hide, drops the bundle and runs `do_get()`. Before the change it printed "You collect 5 low grade hide materials."; after it, "You collect 5 high grade hide materials.", and the high grade count was back to 8. Both versions of `act.item.c` compile without warnings under the build's warning flags.

src/craft/crafting_new.c
moshehbenavraham: Fixed in 94f21665e (Run the supplyorder command from the New Supply Orders special). The special now calls `do_newcraft(ch, argument, cmd, SCMD_NEWCRAFT_SUPPLYORDER)` and returns 1, so the mobile's room gets exactly the command, including its NPC refusal. `supplyorder show` also accepts `info`, which only the special used to handle.

`Test_supply_order_mobile_runs_every_supplyorder_subcommand` calls the special with `materials`, `cancel` and `info` and checks that each reaches its handler. It fails on the old special, and on the new one without `info`.

src/craft/crafting_new.c
moshehbenavraham: Fixed in b5e6224bd (Use block comments on the crafting lines this branch added): the four comments now use `/* */`. A scan of every added line in the branch diff finds no other `//` comment.

moshehbenavraham:
Review addressed. The branch is rebased onto master (638ba31a8) and ready to merge.

Each of the eight review threads has a reply naming its commit and test, and all eight are resolved. Commits added for the review:

- 34e7f7b22 Refuse bonus slot 7 in craft motes
- 6a1b4a55a Let craft score and golem commands run beside a supply order
- f2b89034a Clear a golem project's method when its work ends
- ee5d9df2d Replace a supply offer that has no recipe variant
- 94f21665e Run the supplyorder command from the New Supply Orders special
- 2e6df8313 Offer only the weapon and armor types a recipe builds
- 9555e1ae9 Keep a bundle's hide grade when it is picked up
- b5e6224bd Use block comments on the crafting lines this branch added

Verification at b5e6224bd:

- `make -j16 test` on the host passed: 1657 CuTest tests plus the shell gates.
- Each new or extended test was run against the code without its fix and failed. The `USE_NEW_CRAFTING_SYSTEM` pickup, which is off in the default build, was checked the same way in a scratch build with the option.
- All 30 CI jobs this pull request triggers passed locally with `scripts/ci/local/run.py`, including the whole-tree `quality-clang-tidy` gate. GitHub's checks also finished green (61 passed, 1 skipped).

Please merge with Rebase and merge rather than squash, since the commits are grouped by area. The hashes above will change on merge.

moshehbenavraham:
Merged with Rebase and merge: master moved from a8a6ab71a to 9288eaa2f.

## What merged

All the fixes described above for the materials-and-motes crafting system:

- Boot data: `boot_db()` fills the recipe table and material sort.
- Projects: the edit lock while work runs, the completion recheck, the recipe variant's skill, bonuses on a recipe's first variant, and woodworking without a tool.
- Supply orders: kept apart from crafting projects, abandon refunds unused materials, every new order starts with no progress, all-or-nothing material use, `CrCT` contract terms, and the New Supply Orders special runs the general command.
- Golems, storage and reforge: `CrGo` resume, the chosen wood, hide grades kept by stored bundles, and reforge abbreviations.

The help and docs describe the fixed behavior. The code in `src/`, `unittests/` and `CMakeLists.txt` is identical to the reviewed head b5e6224bd.

## Rebase onto #190

#190 merged after the review round, so the branch was rebased onto a8a6ab71a before merging. #190 had documented the crafting code as it was before these fixes, which caused conflicts in `Makefile.am`, `sql/components/ci_schema_manifest.txt`, `lib/text/help/help.hlp` and `docs/systems/CRAFT_ACTIVITY_LIFECYCLE.md`.

- `help_crafting_project_entries.sql` was dropped. Its text replacements targeted the pre-#190 topics, and #190's `help_crafting_entries.sql` now owns those topics.
- The lifecycle guide conflicts were resolved in each commit that touches it.
- One new commit, 9288eaa2f, aligns 13 help topics with the fixes: CRAFTING, CRAFT-ITEMTYPE, CRAFT-MATERIALS, CRAFTING-RECIPES-MATERIALS, VESSELS-AND-ADVANCED-CRAFTING, NEWCRAFT, CRAFT-SHOW, CRAFT-CHECK, CRAFT-BONUSES, CRAFT-SPECIFIC-TYPE, CRAFT-VARIANT, SUPPLYORDER and REFORGE. The same commit updates the crafting system notes and the OEDIT guide with its generated page. The topics are regenerated in `help_crafting_entries.sql` from `help.hlp`, with new verifier hashes.

## Verification

- Host `make -j16 test` at the rebased head: 1657 CuTest tests plus the shell gates passed.
- Local `scripts/ci/local/run.py` at eb1576b70 (the pushed head, same tree as master now): the ten jobs whose inputs the rebase changed passed. Those are format check, hygiene, gitleaks, database migration, world validation, build parity, both clean-archive builds, world-data tools and the production-linked tests (both I/O drivers). The other 20 jobs passed at b5e6224bd; the C sources, tests and source lists they build did not change.
- `help_crafting_entries.sql` ran twice against temporary copies of the development help tables, and the verifier passed 37/37. It was then applied to the development database, where the verifier also passed 37/37. Before applying, all 18 topics there held either the pre-#190 text or this pull request's earlier edits.
- GitHub's checks on eb1576b70 were still running at merge time; none had failed.

## Commit map

| Reviewed | On master | Commit |
| -- | -- | -- |
| d8304c19e | 119e551d5 | Load crafting recipes and the material sort at boot |
| 2fe7afec2 | b0684d103 | Tighten the crafting project lifecycle and material handling |
| 273698f2a | 6f9e95d6a | Keep supply orders and crafting projects apart |
| 7522fa166 | a9b096e44 | Resume golem construction and keep material grades |
| 1757c7a32 | b0d55637d | Match reforge types by abbreviation and count kit contents |
| 9ada86c87 | 71a41bade | Update the crafting docs for the project fixes (was "Update crafting help and docs for the project fixes"; the help component moved to 9288eaa2f) |
| f79241127 | da91afdfb | Check the crafting recipe table in the world boot test |
| 40fff229d | 0ab908f81 | Parse the new crafting player-file tags with strtol |
| 34e7f7b22 | 37a40bf60 | Refuse bonus slot 7 in craft motes |
| 6a1b4a55a | eaf26d6e2 | Let craft score and golem commands run beside a supply order (lifecycle text adapted) |
| f2b89034a | 71c1a3af3 | Clear a golem project's method when its work ends (lifecycle text adapted) |
| ee5d9df2d | baab04893 | Replace a supply offer that has no recipe variant (lifecycle text adapted) |
| 94f21665e | f10065349 | Run the supplyorder command from the New Supply Orders special |
| 2e6df8313 | bc008cc18 | Offer only the weapon and armor types a recipe builds |
| 9555e1ae9 | f19b56d3c | Keep a bundle's hide grade when it is picked up |
| b5e6224bd | fe2c43a22 | Use block comments on the crafting lines this branch added |
| - | 9288eaa2f | Align the crafting help and notes with the project fixes |

The hashes in the review replies and the summary comment above refer to the "Reviewed" column.

## Remaining follow-up

- `material_grade()` has no grade for brass, cotton, bone and dragonblood.
- The supply reward's material bonus reads allocations after the last item has used them, so it is 0 at completion.
- #92: `Test_supply_order_requests_start_fresh_with_orderable_recipes` and `Test_golem_work_that_ends_leaves_supply_orders_available` in `unittests/CuTest/test_crafting_projects.c` call `circle_srandom(1)` and then reseed from `time(NULL)`. This pull request merged first, so the #92 branch moves both onto its seeding rule, as agreed in the cross-link above.

PR #202, opened 2026-09-17
Expand sanitizer and fuzz coverage across the runtime trust boundaries

### LINKED ISSUE #91: Modernization: expand sanitizer and fuzz coverage across runtime trust boundaries

moshehbenavraham:
Closes #91.

## What changes

**Sanitizer matrix (`test.yml`, job `sanitizers`).** ASan+UBSan on GCC 13 and on Clang 18, and ThreadSanitizer on Clang 18. Each entry sets `CC`, proves the compiler family and version, and runs the new `scripts/ci/check_sanitizer_build.sh`, which prints the configured `CFLAGS`/`LDFLAGS` and the sanitizer runtime each binary links (`__asan_init`, `__ubsan_handle_*`, `__tsan_init`, libFuzzer entry points, static or via the shared runtime) and fails when any is missing. The syntax-check boot now runs inside the instrumented suite (the skip variable is gone from these jobs), the instrumented server is installed and booted through autorun against the isolated MariaDB, `scripts/ci/smoke_client.py` holds a pre-authentication conversation with it (a declined new account name, an invalid name, an over-long line, Telnet negotiation, an empty name that closes the connection), the health endpoint is checked again, and a graceful shutdown is required. `halt_on_error=1` is set for all three sanitizers and there is no suppression file. Logs are uploaded on failure.

**Production-linked fuzz targets (`luminari_fuzz`, job `fuzz`).** A libFuzzer executable that links the cutest objects, so every target drives the parser the server compiles and there is no mirror to drift:

| Target | Boundary |
| -- | -- |
| `dotenv` | `lib/.env` assignments |
| `config` | `lib/etc/config` (`load_config`) |
| `dg` | DG script expressions and `%variable%` substitution |
| `world` | world, mobile, object, zone, and trigger file records |
| `command` | socket line assembly (`process_input`, Telnet stripping) and the command tokenizers |
| `i3` | Intermud3 line framing and JSON-RPC decoding |
| `ai` | AI provider response decoders (OpenAI, Ollama) |
| `discord` | Discord bridge inbound JSON |
| `onboarding` | web onboarding envelopes, capability, and catalog controls |

The world loader and the config reader end the process on a rejected file, so those two run behind an exit guard (`exit()` is interposed and returns to the target) and without leak detection; every other target keeps leak detection on. Seeds, dictionaries, and regression inputs are checked in under `unittests/CuTest/`. `Test_fuzz_targets_replay_seed_and_regression_inputs` replays all of them deterministically in the ordinary production-linked suite in a forked child that fails on a signal, an unexpected exit status, or a sanitizer report. `scripts/ci/run_fuzz_targets.sh` runs each target with its dictionary, keeps and minimizes every reproducer under `fuzz-artifacts/`, and the job uploads that directory on failure. Pull requests get the replay plus 15 s per target and the existing protocol and binary-format fuzzers; `fuzz-campaign.yml` runs longer campaigns weekly and on demand.

**Defects the campaigns found**, each fixed at the root with its reproducer saved as a regression input:

- `asciiflag_conv_aff` shifted a 32-bit `int` past its width for uppercase affect letters from `F` on.
- A mob `Feat`/`MFeat` espec indexed the feat arrays with any number (out-of-bounds write at boot).
- `check_object` read a NULL short description.
- `parse_trigger` tokenized a NULL command list for an empty script.
- `flush_queues` left a live descriptor pointing at a pooled large output buffer, so a second `--` command linked the pool into a cycle and hung the server (player-triggerable).
- `var_subst` stepped past the terminator of an unterminated `%variable` and read beyond its stack buffer when the line filled it.
- Script arithmetic (`* / + -`) overflowed `int`; it now evaluates in `long long`.
- `load_default_config` leaked (and, after a rejected `menu` string, double-freed) its strings when the defaults were loaded again.
- The legacy 3/4-field record conversion indexed `zone_table[-1]` for a record outside every zone.
- `setup_dir` indexed `dir_option[]` with any `D<n>` exit direction; found by the fuzz job's first 15-second run on GitHub, whose uploaded, minimized reproducer became the regression input.
- `matching_quote` skipped the terminator after a trailing backslash in a quoted script token and read past the line (review finding).
- The `--` command discarded a player's pending output, including its own cancellation notice, whenever that output had grown past the small buffer; it now drops only the input queue (review finding).

**Review follow-up on the harness.** The config target now feeds `load_config_stream()` (split out of `load_config()`, which still opens the file) an in-memory stream it owns and closes after a guarded exit, replacing the helper that closed a range of descriptors and left the parser's `FILE` behind. The world target bounds its index tables by every `#` byte, because `get_line()` and `fread_string()` split a line longer than `READ_SIZE - 1` bytes and its tail can become a record header the newline-based count missed; the input that overflowed `parse_room()` through that path is a regression input.

**Documentation.** `docs/guides/TESTING_GUIDE.md` now has a "Sanitizers and Fuzzing" section with the job table, the reproduction commands, the supported compilers cited from the compiler policy, the no-suppression rule, the MemorySanitizer assessment (not enabled: the MariaDB client, json-c, libevent, libcurl, OpenSSL, libgd, and libcrypt are uninstrumented), and the OSS-Fuzz suitability note. The CMake guide documents `LUMINARI_FUZZ`.

## Acceptance criteria from the issue

- ASan+UBSan on both intended compilers, failing on the first unsuppressed finding: the `sanitizers` matrix with `halt_on_error=1` and `check_sanitizer_build.sh`.
- Syntax boot and a pre-auth client interaction under sanitizers: in all three matrix entries.
- TSan covers the threaded subsystems: the full production-linked suite (AI worker, Intermud3 client thread, logging, database handoff) plus boot and client smoke run under `-fsanitize=thread`; nothing is excluded.
- Each named trust boundary has a production-code fuzz target: the nine targets above plus the existing protocol and binary-format fuzzers.
- CI uploads minimized reproducers and logs on failure: `fuzz-artifacts/` and `sanitizer-logs/`.
- Fuzz findings become permanent deterministic tests: the regression replay test plus focused CuTest cases.
- Reproduction commands and supported platforms are documented.

## Verification

- `luminari_fuzz` (Clang 18, `-fsanitize=fuzzer-no-link,address,undefined`): four local campaigns of 6 to 10 minutes across all nine targets on this branch found the defects above; the final campaign on the fixed tree ran clean, and every seed and regression input replays clean.
- Full production-linked suite under ASan+UBSan with the syntax boot against a fresh isolated MariaDB: 1662 tests OK. Under ThreadSanitizer: 1662 tests OK, no reports.
- Every job of `test.yml`, `integration.yml`, `quality.yml`, `hygiene.yml`, and `security.yml` (33 jobs including the three sanitizer entries and the fuzz job) passed as local container replicas (`scripts/ci/local/run.py`) on the review follow-up commits; the two GCC Release jobs, the GCC 16 warning budget, the Debug Clang job, and the clang-tidy gate were rerun on the final commit after the follow-up fixes to the new test code.
- The review findings were reproduced before fixing: the old record count overflowed `parse_room()` under AddressSanitizer on the new world input, and the old `matching_quote` read past the substitution line on the new dg input; both new focused tests fail on the old code and pass on the fixed code.
- GitHub's own runs: all three sanitizer entries passed end to end (1660 tests, the client interaction, and the graceful shutdown), and the fuzz job proved its artifact path by finding the exit-direction defect above and uploading its minimized reproducer.


moshehbenavraham:
## Adversarial Review: PR #202

### Finding 1: Heap Buffer Overflow / OOB Write via `fuzz_count_records` Underestimation

- **Location**: `unittests/CuTest/test_fuzz_targets.c:301-344` (`fuzz_count_records` and buffer allocation in `fuzz_world_run`)
- **Exploit / Repro**:
  `fuzz_count_records(data, size)` estimates record count by scanning for `#` where `index == 0 || data[index - 1] == '\n'`:
  ```c
  if (data[index] == '#' && (index == 0 || data[index - 1] == '\n'))
  ```
  `fuzz_world_run` allocates fixed-size arrays based on this count:
  ```c
  records = fuzz_count_records(data, size) + 2;
  world = calloc(records, sizeof(*world));
  trig_index = calloc(records, sizeof(*trig_index));
  ```
  However, `discrete_load()`, `parse_room()`, `parse_trigger()`, and `parse_simple_mob()` read lines using `get_line()`. `get_line()` reads through `fgets(..., READ_SIZE=256, fl)`.
  If a comment line starting with `*` or any record field exceeds 254 bytes without a newline, `fgets()` consumes 255 bytes. The next call to `get_line()` continues reading from byte 256. If that 256th byte starts with `#<vnum>`, `get_line()` treats it as a valid record header line and passes it to the parser.
  Because byte 255 in the raw input was not `\n`, `fuzz_count_records()` does not count this record. When multiple records follow this pattern, `fuzz_count_records()` returns 0, resulting in `records = 2`. `discrete_load()` proceeds to parse 3+ records and writes out of bounds into heap memory (`trig_index[top_of_trigt++]`, `world[room_nr++]`), triggering an AddressSanitizer heap-buffer-overflow crash.
- **Fix**:
  Do not rely on a naive raw-byte scan to dimension static arrays. Either:
  1. Dynamically reallocate `world` and `trig_index` inside the world loader / test harness when `top_of_trigt >= records` or `room_nr >= records`.
  2. Implement `fuzz_count_records()` using the exact `get_line()` tokenizing logic rather than an unchunked newline search.

---

### Finding 2: Out-Of-Bounds Stack Memory Read in `matching_quote`

- **Location**: `src/dgscript/dg_scripts.c:1898-1910` (`matching_quote`)
- **Exploit / Repro**:
  ```c
  char *matching_quote(char *p) {
    for (p++; *p && (*p != '"'); p++) {
      if (*p == '\\')
        p++;
    }
    if (!*p) p--;
    return p;
  }
  ```
  If an input string begins with `"` and ends with an unescaped trailing backslash before `\0` (for example, `"\\`), `*p == '\\'` evaluates to true and executes `p++`. This advances `p` to point directly to the string null terminator `\0`.
  The loop increment `p++` then executes unconditionally, advancing `p` past `\0` into adjacent unallocated or stack memory. The loop continues evaluating arbitrary memory beyond the buffer until hitting a random `"` or crashing.
  Passing `"\\` to `matching_quote()` immediately produces:
  `AddressSanitizer: stack-buffer-overflow READ of size 1`
  This is reachable via DG script variable evaluation, expression evaluation (`dg_triggers.c:47`), and the `fuzz_dg_run` target.
- **Fix**:
  Ensure the character following the backslash is not the null terminator before advancing:
  ```c
  if (*p == '\\' && *(p + 1))
    p++;
  ```

---

### Finding 3: Arbitrary File Descriptor Destruction and Stale `FILE *` Leak in Config Fuzz Target

- **Location**: `unittests/CuTest/test_fuzz_targets.c:178-218` (`fuzz_close_descriptors_from`)
- **Exploit / Repro**:
  ```c
  static void fuzz_close_descriptors_from(int first) {
    for (int fd = first; fd < first + 16; fd++)
      close(fd);
  }
  ```
  This helper is called after `load_config()` terminates via `longjmp`.
  1. `load_config()` opens a single `FILE *fl`. Blindly calling `close()` on 16 consecutive descriptors starting from `first` closes unrelated descriptors owned by libFuzzer (corpus directory descriptors, coverage output pipes, IPC handles) or the host runtime.
  2. Invoking `close(fd)` directly bypasses `fclose(fl)`. glibc retains the allocated `FILE` object linked in its global `_IO_list_all` list pointing to an invalidated descriptor. Subsequent iterations that open new files get assigned the recycled file descriptor while glibc retains dangling duplicate stream handles.
- **Fix**:
  Track the active `FILE *fl` explicitly and call `fclose(fl)` upon error return/longjmp rather than closing arbitrary integer descriptor ranges.

---

### Finding 4: Pending Output Destruction and Data Loss on `--` Player Command

- **Location**: `src/core/comm.c:3348-3360`, `src/core/comm.c:4280-4284` (`flush_queues` called by `--` in `process_input`)
- **Exploit / Repro**:
  The `--` command is intended to clear a player's pending *input* queue (`"All queued commands cancelled.\r\n"`).
  In `process_input()`:
  ```c
  if (!strcmp(comm, "--")) {
    write_to_q("All queued commands cancelled.\r\n", &t->output, 0);
    flush_queues(t);
    return (1);
  }
  ```
  `flush_queues()` is the descriptor cleanup routine designed for `close_socket()`.
  Inside `flush_queues(t)`:
  - If `t->large_outbuf != NULL` (any queued output $\ge 1024$ bytes, such as a large `who`, `score`, inventory, room descriptions, or combat buffer waiting to be flushed to socket), `large_outbuf` is returned to `bufpool`, `t->bufptr` is reset to 0, and `small_outbuf[0] = '\0'`. All pending server output to the player is silently discarded.
  - If `t->large_outbuf == NULL` (pending output $< 1024$ bytes), `small_outbuf` is not cleared, so small outputs survive.
  A player executing `--` silently destroys pending game output whenever the output buffer exceeds 1 KB.
- **Fix**:
  `--` in `process_input()` should only flush `t->input`, freeing pending `txt_block` entries and resetting `t->input.head = t->input.tail = NULL`. It should not call `flush_queues(t)` and must not touch `t->output` or `t->large_outbuf`.

moshehbenavraham:
All four findings are addressed on the branch; each was reproduced first.

**Finding 1 (world record count).** Confirmed. `READ_SIZE` is 512 rather than 256, so the split falls at 511 bytes, but the mechanism is exactly as described: a comment chunk of 511 bytes with no newline followed by `#N` produced a header the newline-based count never saw, and the third such record wrote past `world[]` (AddressSanitizer heap-buffer-overflow in `parse_room`, reproduced with the old count). The count now takes every `#` byte as the bound, which covers any split however the readers chunk the input (`fread_string` reads with the same `READ_SIZE - 1` limit, so simulating `get_line` alone would still have undercounted). The input is saved as `fuzz_regressions/world/wld-headers-after-split-comment-lines.txt`.

**Finding 2 (`matching_quote`).** Confirmed. `"\` followed by the terminator advanced past it; on a substitution line that filled the buffer this was a stack-buffer-overflow read under AddressSanitizer through the `dg` target. The escape is now skipped only when a character follows the backslash. `Test_dg_production_matching_quote_stops_at_a_trailing_backslash` covers the trailing, escaped, and unterminated cases (it fails on the old code), and `fuzz_regressions/dg/quoted-trailing-backslash.txt` fills the line so the sanitizer sees the read.

**Finding 3 (descriptor closing).** Fixed by removing the mechanism rather than narrowing it. `load_config()` now opens the file and hands it to `load_config_stream()`, which loads the defaults and reads the settings from any open stream. The config target owns an `fmemopen` stream and closes it after the guarded exit, the same way the world target already did, so `fuzz_next_descriptor` and `fuzz_close_descriptors_from` are gone along with the scratch config file. The defaults are still reloaded on every iteration, so the leak and double-free regression input keeps its meaning. A 25-second run of the config target completed 208k iterations with four descriptors open mid-run.

**Finding 4 (`--` discarding output).** Confirmed and fixed. The command now drops only the input queue through a shared `flush_input_queue()`; `flush_queues()` uses the same helper and still resets a reused descriptor to its own buffer, which the fuzz targets rely on. `Test_cancel_queued_commands_keeps_pending_output` queues more output than the small buffer holds plus a command, sends `--`, and checks the output and the cancellation notice survive while the input queue is empty (it fails on the old code).

Verification on the final tree: the full production-linked suite (1662 tests) with the syntax boot, all seed and regression inputs replayed through `run_fuzz_targets.sh --seconds 0` under ASan+UBSan, and all 33 local CI replica jobs (the five affected by the follow-up test code were rerun on the final commit). The testing guide's defect list is updated.

moshehbenavraham:
Merged to master by rebase as 15 commits (6f8363111 through e6efd9ddf).

**What landed**

- The `sanitizers` matrix in `test.yml`: ASan+UBSan on GCC 13 and Clang 18, ThreadSanitizer on Clang 18, each proving the linked runtime, running the full production-linked suite and the syntax boot instrumented, booting the instrumented server through autorun against the isolated MariaDB, and holding a pre-authentication client conversation. `halt_on_error=1`, no suppression file, logs uploaded on failure.
- `luminari_fuzz`, the production-linked libFuzzer harness with nine targets (dotenv, config, dg, world, command, i3, ai, discord, onboarding), seeds, dictionaries, and regression inputs under `unittests/CuTest/`, replayed deterministically in the ordinary suite. The `fuzz` job runs 15 s per target on pull requests and uploads minimized reproducers; `fuzz-campaign.yml` runs longer campaigns weekly and on demand.
- Twelve defects found by the campaigns and the review, each fixed at the root with a saved reproducer, including two player-reachable server issues in the `--` command (a hang from a dangling large output buffer, and discarded pending output).
- The "Sanitizers and Fuzzing" section of `docs/guides/TESTING_GUIDE.md`, with reproduction commands, supported compilers, the no-suppression rule, and the MemorySanitizer and OSS-Fuzz assessments.

**Follow-up work**

- None required for #91. Two items remain deliberately out of scope and are documented in the guide: MemorySanitizer stays off until the dependency set (MariaDB client, json-c, libevent, libcurl, OpenSSL, libgd, libcrypt) can be built instrumented, and an OSS-Fuzz project would need a build script providing those shared libraries and a target-per-binary split of `luminari_fuzz`.
- The weekly `fuzz-campaign.yml` run is the ongoing source of new findings; any reproducer it uploads should be fixed at the root and saved under `fuzz_regressions/<target>/` the same way.

PR #203, opened 2026-09-17
Raise risk-weighted test coverage and remove randomized coverage variance

### LINKED ISSUE #92: Modernization: raise risk-weighted test coverage and eliminate randomized CI variance

moshehbenavraham:
## Summary

Closes #92.

Rerunning the Coverage job at one commit now measures identical totals. Every coverage floor is a
measured value that can only rise. Critical subsystems and the lines a change touches have their
own floors, and the gaps in the eight critical areas the issue names have tests. The new tests also
found two production defects, which this PR fixes.

## Why the totals moved

- Tests drew random numbers from whatever state earlier tests left: 26 clock reseeds in test files,
  and a generator that returns one constant until something seeds it. Two Coverage runs of
  638ba31a8 differed by 9 lines and 13 branches.
- The supply-order rules read the wall clock: offer recipes come from an hourly seed, and event
  contracts appear on Sundays.
- The domain-event world registry and event bus hash entity addresses. Their collision paths ran
  only when the address layout put two entities in one bucket, about one run in six for the
  registry.
- Event profile reports sort on measured callback time, so their tie-break paths ran only when two
  timings matched.
- The environment file cache compares a cached file's identity field by field. After a same-size
  edit, the first field that differs depends on whether a file clock tick fell between two writes.

## Changes

1. **Seed every production-linked test from one recorded seed.** The generated runner reads
   `LUMINARI_TEST_SEED` (default 1) and prints it. Before each test it seeds the game and C library
   generators from the seed and the test name, so `LUMINARI_TEST_SEED=<n> CUTEST_FILTER=<test>`
   replays exactly what the test drew in the full run. A failing run prints that command.
   `make test` and CTest pin the default seed, and `make test-seeds SEEDS="..."` runs the suite once
   per seed. Tests no longer reseed from the clock, and `test_cutest_runner.py` rejects a test that
   does. `init_perlin` now reseeds the C library generator from the game generator instead of the
   clock. Forked test children exit through `CuTestChildExit`, which writes their coverage.
2. **Keep coverage independent of the calendar, address layout, and timing.** In the test build,
   the supply-order rules read one pinned instant. The Coverage job runs the suite without address
   randomization and pads its environment to one size, and the local CI runner allows that
   personality change. New tests force a world registry collision and a perfmon report tie on
   purpose, and change each identity field the environment file cache compares.
3. **Stop `one_word` at an unterminated quote and cover alias expansion.**
4. **Read shop price factors as doubles and round-trip zone and shop files.**
5. **Test the lifecycles critical subsystems left unexecuted:** account character removal commit
   and rollback (MariaDB), the descriptor lifecycle over a loopback connection, the I3 client
   thread's handoff and shutdown, and password buffer bounds. Each step of the I3 test waits for
   an effect of the step before it, so the thread loop takes the same paths however the threads
   are scheduled.
6. **Hold coverage and mutation scores to measured floors.** `scripts/ci/coverage_policy.json` and
   `scripts/ci/check_coverage.py` add repository, subsystem, and changed-line floors. The checker
   also requires normalized `src/` report paths, and named tests and owners that exist.
   `check_baseline_ratchet.py` refuses a change that lowers or removes a floor. The Coverage job
   compares with `HEAD^1` and keeps `coverage-policy.txt` and the HTML report's script. It uploads
   the artifacts even when a check fails. `scripts/ci/mutation_test.py` measures assertion quality
   for the binary file decoders, the telnet/MSDP/GMCP input parsers, and password hashing. The
   weekly and manual Extended Tests workflow runs seed stress and mutation testing.
7. **Exchange a login with the server in the startup smoke test.** Over telnet, the test creates an
   account and checks the isolated database row, which must hold a current-scheme hash. It then
   logs in again after one rejected password and fails on any `SYSERR`.
8. **Document** seeds and replay, seed stress, the coverage policy and critical subsystems,
   mutation testing, the test profiles, and the smoke test exchange.

## Defects found and fixed

- **Shop prices.** #185 widened `profit_buy`/`profit_sell` to `double`, but `boot_the_shops` still
  read them with `"%f"`. Every shop loaded from a world file got buy and sell factors near zero.
  The new zone/shop writer-to-loader round trip found it.
- **`one_word` overrun.** An unterminated quote returned a pointer one byte past the string
  terminator. Help keyword lines and the DG `skillset` field call it on data-file text.
- **`Test_artifact_integration_signature_procs_run_without_a_roll`** passed only with the constant
  generator. The weighted shape does nothing on 30 percent of its rolls by design.

## Acceptance criteria

| Criterion | Evidence |
| -- | -- |
| Reruns produce the same seed and stable totals | Every run prints `LUMINARI_TEST_SEED=1`. At 52cef6aa2, local CI Coverage runs on 1, 2, and 16 CPUs and the GitHub Coverage run all measured 86427/275573 lines and 50604/253223 branches, with no line or branch differing. Earlier runs with four different environment sizes also matched. |
| Critical code cannot lower its subsystem or changed-line threshold | Eight subsystem floors plus changed-line floors. In this PR, every changed line in a critical subsystem is covered: 2 of 2 in command parsing, 4 of 4 in networking, and 2 of 2 in the world/DG/config parsers. |
| Repository floors only increase | Floors rise to 31.36% lines and 19.98% branches (from 10.50 and 7.16). The ratchet check refuses lower or removed floors. |
| Debug, optimized, sanitizer, and coverage profiles run the suite | CMake Debug and Release (`-O3`) on four compilers, the hardened `-O2` production profile, ASan/UBSan, Valgrind, and Coverage. See Test profiles in the testing guide. |
| Critical gaps have named tests and owners | Each subsystem lists its sources, owner test file, and named tests. The checker validates the tests against `cutest_test_files`. |
| CI retains human- and machine-readable artifacts | HTML with CSS and JS, Cobertura XML, JSON summary, and `coverage-policy.txt` (`if: always()`), plus `mutation-report.txt`. |
| A documented stress job replays any failing seed | The Extended Tests `seeds` input, `make test-seeds`, and the replay commands in the testing guide. |

Measured floors: authentication 63.82/51.17, command parsing 26.39/26.03, SQL 43.10/35.15,
world/DG/config parsers 36.87/28.81, persistence 61.37/43.43, networking 44.75/27.89, OLC
29.74/26.43, threaded services 41.61/27.91 (lines/branches). Mutation scores: binary formats 71.91,
protocol parsers 32.08, password 88.64 (floors 71.91, 32.07, and 88.63). Each report lists its
surviving mutants.

## Verification

- Review fixes at ef4f99df6 (macro masking, padding clamp, reset handling, forked I3 test):
  local CI 30 of 30 jobs passed, Coverage measured the same totals and floors as 52cef6aa2,
  seeds 1 through 5 passed, and mutation scores are unchanged.
- Local CI, `python3 scripts/ci/local/run.py --jobs 4 --cpus 4` at 52cef6aa2: all 30 jobs passed,
  including Coverage, clang-tidy, CMake Debug and Release on four compilers, the production
  profiles, ASan/UBSan, Memory check, and both clean archives. The unit-test job ran all 1667 tests
  with both I/O drivers, then the smoke test's login exchange.
- GitHub Build & Test on the pushed branch at 52cef6aa2: 24 of 24 jobs passed.
- Coverage determinism at 52cef6aa2: `run.py --job test-coverage` on 1, 2, and 16 CPUs, the
  matrix's Coverage job, and GitHub's Coverage run produced Cobertura reports with no line or
  branch differing.
- Seeds: `LUMINARI_TEST_SEED=<n> ./cutest` passed for seeds 2 through 9 at 52cef6aa2, and for seeds 1
  through 90 earlier in the branch.
- `python3 scripts/ci/mutation_test.py` measured the same scores on the host and in the CI image.
- `python3 scripts/ci/check_baseline_ratchet.py --base origin/master`: no baseline grew.

## Notes for reviewers

- The first GitHub Coverage run of this branch measured 1 line and 3 branches fewer than local CI.
  GitHub's runner scheduled the I3 thread test differently (1 line and 4 branches fewer), and a
  file clock tick fell inside the environment cache test (1 branch more), which a loaded local run
  then reproduced. Commits 2 and 5 now make both deterministic, and the next GitHub run matched
  local CI exactly.
- The changed-line floor applies to the eight critical subsystems only. Code that no unit test can
  reach, such as `game_loop`, needs a test hook or a restructure before it changes.
- When a PR raises coverage, the Coverage job prints the new floor values to record in the policy.
- Hosted runners allow the Coverage job's `setarch --addr-no-randomize` step. The local runner now
  starts containers with `seccomp=unconfined`, because Docker's default profile refuses that
  personality change.


moshehbenavraham:
### Adversarial Review Findings

#### 1. [P1 Correctness / AST Integrity] Multiline macro masking drops newlines, corrupting mutant line calculations
- **Location**: `scripts/ci/mutation_test.py:163-168`
- **Issue**:
  In `mask_non_code()`, comments and string literals preserve newline characters (`if text[position] != "\n": masked[position] = " "`) so that offsets and line counts remain synchronized with the original text. However, preprocessor directive masking at lines 166-167 replaces all characters with spaces unconditionally:
  ```python
  while end < length and not (text[end] == "\n" and text[end - 1] != "\\"):
      end += 1
  for position in range(index, end):
      masked[position] = " "
  ```
  When a macro spans multiple lines via backslash continuations (`\\\n`), the embedded newlines are overwritten with spaces. This drops newlines in `masked`, desynchronizing line numbers and causing `self_test()`'s invariant (`assert masked.count("\n") == source.count("\n")`) to fail on any file containing multiline macros.
- **Repro**:
  ```python
  from scripts.ci.mutation_test import mask_non_code
  s = "#define FOO(a, \\\n           b) \\\n  ((a) + (b))\nint bar(void) { return 1; }\n"
  assert s.count("\n") == mask_non_code(s).count("\n")
  # AssertionError: Newline count mismatch (Original: 4, Masked: 2)
  ```
- **Fix**:
  Preserve newlines when replacing characters in preprocessor blocks:
  ```python
  for position in range(index, end):
      if text[position] != "\n":
          masked[position] = " "
  ```

---

#### 2. [P1 CI Stability / Hang] Negative environment padding causes `head -c` to hang infinitely on `/dev/zero`
- **Location**: `.github/workflows/test.yml:730-732`
- **Issue**:
  ```bash
  padding=$((65536 - $(env | wc -c)))
  COVERAGE_ENV_PADDING=$(head -c "$padding" /dev/zero | tr '\0' x) \
    setarch "$(uname -m)" --addr-no-randomize ./cutest
  ```
  If the runner's exported environment exceeds 64KB (65,536 bytes) — e.g. runners with large token variables, workflow matrix inputs, secrets, runner diagnostics, or composite actions — `65536 - $(env | wc -c)` yields a negative integer `-N`.
  In GNU `head`, `-c -N` instructs `head` to output "all but the last N bytes". When reading from `/dev/zero` (an infinite byte stream that never produces EOF), `head` loops infinitely attempting to reach the end of the stream, hanging CI until runner timeout.
- **Repro**:
  ```bash
  head -c -5 /dev/zero  # Infinite loop / hang
  ```
- **Fix**:
  Clamp padding to a minimum of 0:
  ```bash
  padding=$((65536 - $(env | wc -c)))
  if [ "$padding" -gt 0 ]; then
    COVERAGE_ENV_PADDING=$(head -c "$padding" /dev/zero | tr '\0' x)
  else
    COVERAGE_ENV_PADDING=""
  fi
  COVERAGE_ENV_PADDING="$COVERAGE_ENV_PADDING" setarch "$(uname -m)" --addr-no-randomize ./cutest
  ```

---

#### 3. [P2 CI Robustness] Unhandled `ConnectionResetError` crashes smoke client with traceback instead of diagnostic exit
- **Location**: `scripts/ci/smoke_client.py:73, 88`
- **Issue**:
  In `Session.expect()` and `Session.expect_close()`:
  ```python
  def expect_close(self):
      deadline = time.monotonic() + self.timeout
      while time.monotonic() < deadline:
          self.sock.settimeout(max(deadline - time.monotonic(), 0.01))
          try:
              if not self.sock.recv(4096):
                  self.sock.close()
                  return
          except socket.timeout:
              continue
      raise SystemExit("the server did not close the connection after Quit")
  ```
  If the MUD server closes or resets the TCP connection abruptly (e.g. TCP RST on shutdown or unread buffer discard), `self.sock.recv(4096)` raises `ConnectionResetError` (subclass of `ConnectionError`/`OSError`).
  Because `expect_close()` only catches `socket.timeout`, a TCP RST causes a raw Python exception and stack trace rather than recognizing that the connection closed. Similarly, if RST occurs during `expect()`, it raises an uncaught exception rather than the formatted diagnostic `SystemExit(f"connection closed before {marker!r}...")`.
- **Repro**:
  Run `expect_close()` against a server socket that issues `SO_LINGER` timeout 0 or terminates with pending unread data. `smoke_client.py` fails with:
  `ConnectionResetError: [Errno 104] Connection reset by peer`
- **Fix**:
  Catch `(ConnectionError, OSError)` in both `expect()` and `expect_close()`:
  ```python
  # in expect_close:
  try:
      if not self.sock.recv(4096):
          self.sock.close()
          return
  except (ConnectionError, OSError):
      self.sock.close()
      return
  except socket.timeout:
      continue

  # in expect:
  try:
      chunk = self.sock.recv(4096)
  except socket.timeout:
      continue
  except (ConnectionError, OSError) as error:
      raise SystemExit(f"connection reset before {marker!r}: {error}; received:\n{text}")
  ```

---

#### 4. [P2 Test Isolation / Shared State] Process-wide `chdir()` in unisolated unit test corrupts runner working directory on failure
- **Location**: `unittests/CuTest/test_i3_client_production.c:663-667`
- **Issue**:
  ```c
  if (prepared && chdir(sandbox) == 0)
  {
    started = i3_initialize();
    if (chdir(original_cwd) != 0)
      started = -1;
  }
  ```
  `Test_i3_client_thread_hands_off_both_ways_and_joins_on_shutdown` modifies the process-wide current working directory in the main test runner process.
  Unlike `Test_olc_zone_and_shop_files_round_trip_through_the_loaders` (which isolates `chdir` by running inside a `fork()`ed child process via `CuTestChildExit`), if `i3_initialize()` aborts, crashes, triggers a signal, or if `chdir(original_cwd)` fails, the runner's CWD remains locked to `/tmp/i3-thread-XXXXXX`. All subsequent tests attempting to access relative paths (`lib/`, `world/`, etc.) fail with cascading false positives.
- **Repro**:
  Introduce any failure or assertion inside `i3_initialize()` before `chdir(original_cwd)`. All subsequent file-dependent tests across the suite fail.
- **Fix**:
  Run the test in a forked child process using `fork()` and `CuTestChildExit()` (consistent with `test_world_loading_production.c`), or pass an explicit configuration path to `i3_initialize()` rather than changing process-wide working directory.

moshehbenavraham:
All four findings are addressed in ef4f99df6.

1. **Multiline macro masking.** `mask_non_code` keeps the newline inside a backslash-continued directive, the same way it does for comments and literals. The self-test source now contains a two-continuation macro, and its line and newline-count assertions cover it. No current mutation target has a multiline macro, so the measured scores are unchanged: 71.91 / 32.08 / 88.64.
2. **Negative padding.** The Coverage step clamps the padding at zero, so an environment over 64 KiB runs unpadded instead of feeding `head -c -N` an endless `/dev/zero`. Checked with a 70,000-byte variable exported: padding is 0 and the step continues.
3. **Reset on `recv`.** `expect` turns a reset into the same `SystemExit` diagnostic as an orderly close, and `expect_close` treats a reset as the close it was waiting for. `socket.timeout` is itself an `OSError`, so the timeout clause stays first. Reproduced against a loopback server closing with `SO_LINGER` zero: `expect` reports `connection reset before 'never': [Errno 104] ...`, and `expect_close` returns.
4. **`chdir` in the runner.** The I3 thread exchange runs in a forked child that exits through `CuTestChildExit`, following `test_world_loading_production.c`. The child's exit code names the first failed step (1 sandbox, 2 `i3_initialize`, 3 connect, 4 event fd, 5 event, 6 tell, 7 free, 8 close), and the runner's working directory is never changed. The child still records its coverage, and the Coverage job at ef4f99df6 measured the same totals as before: 86427/275573 lines and 50604/253223 branches, with every floor unchanged.

Verification at ef4f99df6:

- Local CI, `scripts/ci/local/run.py --jobs 4 --cpus 4`: all 30 jobs passed, including Coverage, clang-tidy, ASan/UBSan, Memory check, and the unit-test job's login smoke test.
- `./cutest` on seeds 1 through 5: 1667 tests OK on each.
- `scripts/ci/mutation_test.py`: scores as above, every floor met.
- `scripts/ci/mutation_test.py --self-test` and `actionlint` on the workflow: pass.

The branch is 9 commits ahead of master and 0 behind.

moshehbenavraham:
Merged into master by rebase, so every commit has a new hash:

| Branch | Master | Subject |
| -- | -- | -- |
| 25009e3f2 | 75897929f | Seed every production-linked test from one recorded seed |
| 46073cd8f | 45e98345b | Keep coverage independent of the calendar, address layout, and timing |
| 7de6e0911 | 791f9f7ea | Stop one_word at an unterminated quote and cover alias expansion |
| be1419b9d | 1a4560101 | Read shop price factors as doubles and round-trip zone and shop files |
| babe53ba5 | 44f1f21f6 | Test the lifecycles critical subsystems left unexecuted |
| 76cd7b86e | add621722 | Hold coverage and mutation scores to measured floors |
| b186a66b8 | 69982e1bd | Exchange a login with the server in the startup smoke test |
| be44556ae | 31fef09f7 | Document test seeds, the coverage policy, and mutation testing |
| fb67ec5d8 | 2e17b5c6d | Address review findings on coverage determinism and test isolation |
| 2b16d29c0 | 21382a04a | Decode provider responses through the AI service test wrappers |

**Rebase onto the fuzz-campaign commits (e6efd9ddf).** Two conflicts were resolved by keeping both sides:

- `src/core/comm.c` and `comm.h` both grew test hooks in the same block; all four remain.
- Master's `scripts/ci/smoke_client.py` (a stranger's pre-authentication conversation, added for the sanitizer job) and this branch's file of the same name (account creation and login) are different programs. Master's keeps its name; this branch's is now `scripts/ci/login_client.py`, and `test_server_startup.sh` runs the stranger conversation first, then the login exchange. The testing guide names both.

**One new commit after the rebase.** The fuzz campaign added two CuTest-only wrappers around the AI provider response decoders that only the fuzz harness calls, which put four unexecuted lines into the threaded services subsystem and dropped it to 41.55 percent against its 41.61 floor. `Test_ai_provider_response_decoders_unescape_the_text_field` decodes one OpenAI-style and one Ollama body through them and checks the unescaped text; the subsystem now measures 45.37 percent of lines and 30.84 percent of branches.

**Verification at 2b16d29c0** (identical content to master 21382a04a): local CI matrix 33 of 33 jobs passed, including Coverage (every floor met, repository at 31.40 percent lines and 20.01 percent branches), the new fuzz job, clang-tidy, ASan/UBSan, Memory check, and the unit-test job's startup script with both client exchanges. The full suite passed on seeds 1 through 3 (1673 tests). The baseline ratchet reports no baseline grew. Mutation targets are unchanged by the rebase, so the scores measured earlier stand.

**Follow-up.** The Coverage job reports the repository floors can rise to 31.40 lines and 20.01 branches; the next PR that touches `scripts/ci/coverage_policy.json` should record them. Issue #92 is closed by this merge.

PR #205, opened 2026-09-18
Match help aliases in the database lookup and suggestions

### LINKED ISSUE #204: Help lookup ignores alternate_keywords: 1,449 alias tokens unreachable in game
## Problem

The in-game help lookup reads only the `help_keywords` table. `help_entries.alternate_keywords` holds space-separated alias tokens that the help-sync engine projects onto the keyword line of `lib/text/help/help.hlp`, so the tracked file advertises them, but `help <alias>` fails against the database.

Measured on the development catalog on 2026-09-18 (identical to production after the sync):

- 588 entries have a non-empty `alternate_keywords`
- 415 of them carry 1,449 tokens with no matching `help_keywords` row
- Example misses from `log/help` on production: `help logs` (alias of SYSLOG), `help ammo`, `help archery`, `help bows` (aliases of BLAST)

The file fallback never runs while the database is up, so these only work if MySQL is down.

## Where

- `search_help()` in `src/core/help.c`: the prepared query joins `help_keywords hk` and matches `LOWER(hk.keyword) LIKE LOWER(?)` only.
- `soundex_search_help_keywords()` and `edit_distance_search_help_keywords()` in the same file: both select from `help_keywords` only, so aliases are never suggested either.
- `scripts/help-sync/catalog.py` `legacy_entries_for_catalog()`: projects tag + keywords + aliases into the file, which is why the file and the database disagree on what is reachable.

## Caveat before fixing

Some alias data is misfiled from the old flat-file import, for example GOSSIP, GRATS and CONGRATULATIONS sit on the AUCTION entry and AMMO, ARCHERY, BOWS, QUIVERS sit on BLAST. Promoting every alias blindly would route those lookups to the wrong page, and some would collide with entries that already own the keyword.

## Options

1. Decide that aliases are lookup keywords. Review the 415 entries, then promote the reviewed tokens into `help_keywords` through the normal help editing path and publish with the help-sync engine. No code change; the file projection does not change because it already lists them.
2. Keep aliases as a separate field and teach the three queries to match alias tokens, for example `CONCAT(' ', he.alternate_keywords, ' ') LIKE CONCAT('% ', ?, ' %')`, with the same ordering and level rules. Misfiled aliases would then need cleaning as their own data task.

Either way, add a CuTest that fails while an alias-only token is unreachable, and re-check `log/help` on production after publishing.

## Context

Triage of the production help miss log: `docs/ongoing-projects/production-help-log.md`.

moshehbenavraham:
## Summary

Fixes #204: `help <alias>` failed against the database for the 1,449 alias tokens that `help.hlp` already lists.

- All three lookup queries in `src/core/help.c` (`search_help()`, soundex, edit distance) share a new `HELP_LOOKUP_KEYWORDS_SQL` prefix: a recursive CTE that splits `help_entries.alternate_keywords` into tokens and unions them with `help_keywords`. No schema change, no data migration.
- Ranking: a token equal to the typed word sorts before a keyword that merely starts with it (`help logs` shows SYSLOG, not LOGSEARCH; `help spell-resist` reaches MAGIC-RESISTANCE). When an alias collides with a keyword another entry owns, the keyword owner sorts first (`help gossip` still lists CHAT and AUCTALK before the AUCTION alias). Misfiled aliases from the old flat-file import remain a help data cleanup, as the issue's caveat asks.
- The displayed keyword line includes aliases, so `help logs` is an exact match and does not fall through to the suggestion handler.
- The suggestion merge moved into `help_lookup_suggestions()` so the miss path is reachable from a test.
- New `unittests/CuTest/test_help_alias_lookup.c` shadows the help tables with TEMPORARY copies on the `LUMINARI_TEST_MYSQL_*` database and covers the direct lookup, a prefix-keyword neighbor, the level gate, the collision order, and both suggestion queries. It fails on the old queries.
- Docs: `docs/systems/HELP_SYSTEM.md` describes the lookup keyword set; the production help log triage marks the alias gap done in code.

## Verification

- `make test-all` against a prepared CI runtime with database tests enabled.
- The new test fails on the pre-change queries (rebuilt with the old `help.c` to check).
- SQL prototyped on the development catalog: `help logs` at level 31 returns SYSLOG with `Logs` on its keyword line; the CTE over the whole catalog runs in under 20 ms.

Closes #204


moshehbenavraham:
Local CI replica on the final commit (e4f810c08), all 33 jobs of the matrix:

- 32 pass, including `test-unit-tests`, every `test-cmake-*` and `test-warning-budget-*` entry, both sanitizer jobs, `test-memory-check`, `quality-clang-tidy`, `quality-format-check`, and `hygiene-hygiene`.
- `test-fuzz` reported a stack-buffer-overflow in `strip_colors()` reached from `check_object()` in the world loader. Neither function is in this diff; it is filed as #206 with the sanitizer trace.

Host run: `make test-all` against a prepared CI runtime with the database tests enabled is 1679/1679, and the new alias test fails when rebuilt against the pre-change queries.

src/core/help.c
```diff
@@ -387,17 +408,17 @@ struct help_entry_list *search_help(const char *argument, int level)
...
                                 "LCASE(SUBSTRING(hk2.keyword, 2))) SEPARATOR ', ') "
                                 "FROM help_entries he "
-                                "INNER JOIN help_keywords hk ON he.tag = hk.help_tag "
-                                "LEFT JOIN help_keywords hk2 ON he.tag = hk2.help_tag "
+                                "INNER JOIN help_lookup_keywords hk ON he.tag = hk.help_tag "
+                                "LEFT JOIN help_lookup_keywords hk2 ON he.tag = hk2.help_tag "
                                 "WHERE LOWER(hk.keyword) LIKE LOWER(?) AND he.min_level <= ? "
                                 "GROUP BY he.tag, he.entry, he.min_level, he.last_updated "
-                                "ORDER BY LENGTH(hk.keyword) ASC"))
+                                "ORDER BY MIN(hk.is_alias) ASC, MIN(LENGTH(hk.keyword)) ASC"))
```
moshehbenavraham: **[bug] Exact alias loses to a keyword prefix; only the first row is shown.**

`ORDER BY MIN(hk.is_alias) ASC` ranks any matching `help_keywords` row ahead of an exact alias, including a longer prefix (`LOGSEARCH` vs alias `LOGS`). `handle_database_help` exact-matches and displays only the list head (`help.c:1165-1214`, `1280-1294`), so the alias entry is never shown.

**Repro (shipped CTE + this SELECT, `PREPARE`/`EXECUTE`, TEMPORARY tables):** `syslog` alias `LOGS` plus keyword `LOGSEARCH`; pattern `logs%` at level 31 returns `[0] logsearch` (`min_is_alias=0`, kwline `Logsearch`), `[1] syslog` (`min_is_alias=1`, kwline includes `Logs`). Live catalog: `help spell-resist` returns `resistance` / `Spell-resistance` first; exact alias `Spell-resist` on `magic-resistance` is third. `help logs` itself is fine today only because nothing else matches `logs%`. The new CuTest has no prefix-keyword neighbor, so it misses this.

**Fix:** bind the unprefixed argument and rank exact equality first:
```sql
ORDER BY MIN(CASE WHEN LOWER(hk.keyword) = LOWER(?) THEN 0 ELSE 1 END) ASC,
         MIN(hk.is_alias) ASC,
         MIN(LENGTH(hk.keyword)) ASC
```
`help gossip` still ties on exactness and uses `is_alias` to keep CHAT/AUCTALK ahead of AUCTION.

moshehbenavraham (commented):
## Adversarial review (head e4f810c)

Hunted the shipped lookup path only: `HELP_LOOKUP_KEYWORDS_SQL`, the three queries, `help_lookup_suggestions()`, cache (argument+level), and the TEMPORARY-table CuTest. Prepared-statement binds match (`search_help` 2, soundex 9, edit-distance 3; CTE has zero `?`). `he.min_level <= ?` holds on lookup and both suggestion queries; `help logs` at level 1 is empty; cache is keyed by level. Lookup SQL is SELECT-only. The new CuTest passes and does not restore `conn` on a failing assert, but TEMPORARY shadows mean that is test isolation, not help-table data loss.

GitHub rejects `REQUEST_CHANGES` on this account's own PR, so this is submitted as `COMMENT`. The finding still blocks merge from a review standpoint until the ORDER BY is fixed.

One defect survived.

### Finding 1 — exact alias loses to a keyword prefix; first row is the only one shown

**Location:** `src/core/help.c:421` (`ORDER BY MIN(hk.is_alias) ASC, MIN(LENGTH(hk.keyword)) ASC`) together with `handle_database_help` (`src/core/help.c:1165-1214`, `1280-1294`), which exact-matches and `page_string`s only `entries` (the head).

**Exploit / repro:** `search_help` binds `argument + "%"`. `MIN(hk.is_alias)` is computed over every *prefix* match, so a real keyword that merely *starts with* the typed alias outranks an exact alias token. The player is then shown that first row. If its tag/keyword line is not an exact match, the chain treats it as a partial hit and never displays the alias entry.

Isolated TEMPORARY tables, shipped CTE + `search_help` SELECT, `PREPARE`/`EXECUTE` with `?`:

```sql
-- syslog.alternate_keywords = 'SYSLOG-OFF LOGS AUTOLOG' (min_level 31)
-- logsearch keyword LOGSEARCH (min_level 1)
-- pattern 'logs%', level 31
-- [0] logsearch  min_is_alias=0 min_len=9  kwline='Logsearch'
-- [1] syslog     min_is_alias=1 min_len=4  kwline='Autolog, Logs, Syslog, Syslog-off'
```

`help logs` is supposed to be syslog. The first row is logsearch. `handle_database_help` compares `"logs"` to tag `logsearch` / `"Logsearch"`, calls it a partial match, and prints the wrong entry.

On the current catalog, `help logs` happens to work (nothing else matches `logs%`). A live alias that *does* collide:

```
help spell-resist  (pattern spell-resist%, level 1)
  [0] resistance        kwline='Resistance, Spell-resistance'   -- not exact
  [1] spell-resistance  kwline='Spell-resistance'
  [2] magic-resistance  kwline='Magic-resist, Magic-resistance, Spell-resist'
```

There is no tag or `help_keywords` row `spell-resist`. The exact alias sits on `magic-resistance` and is never shown. Same shape: `help nocharm`, `help trigedit-commandlist`.

`help gossip` still ranks CHAT/AUCTALK ahead of AUCTION because that collision is the *same* token (is_alias 0 vs 1, equal length). The new test only covers that case plus isolated `logs` with no prefix-keyword neighbor, so it stays green.

**Fix:** Rank exact token equality first, then keyword-over-alias, then length. Bind the unprefixed argument as an extra parameter:

```sql
ORDER BY MIN(CASE WHEN LOWER(hk.keyword) = LOWER(?) THEN 0 ELSE 1 END) ASC,
         MIN(hk.is_alias) ASC,
         MIN(LENGTH(hk.keyword)) ASC
```

That keeps `help gossip` (all three rows exact; is_alias still breaks the tie) and makes `help logs` / `help spell-resist` select the exact alias row so the existing first-row exact-match check fires. Extend `Test_help_alias_tokens_are_lookup_keywords` with a `LOGSEARCH` keyword beside the `LOGS` alias and assert the first tag is `syslog`.

moshehbenavraham:
## Review finding 1 addressed (head 72a972fa8)

`search_help` now orders by exact token equality first, then keyword-over-alias, then length, with the unprefixed argument bound as the third parameter:

```sql
ORDER BY MIN(CASE WHEN LOWER(hk.keyword) = LOWER(?) THEN 0 ELSE 1 END) ASC,
         MIN(hk.is_alias) ASC, MIN(LENGTH(hk.keyword)) ASC
```

`Test_help_alias_tokens_are_lookup_keywords` gained a level-1 `LOGSEARCH` keyword beside the `LOGS` alias. It asserts `syslog` first and `logsearch` second at level 31, and only `logsearch` at level 1. Rebuilt against the previous ORDER BY it fails with `expected <syslog> but was <logsearch>`, which is the review's repro.

Read-only check of the shipped CTE plus the new ORDER BY on the development catalog, first row per lookup:

| Lookup | Level | First row |
| -- | -- | -- |
| `help spell-resist` | 1 | magic-resistance (exact alias) |
| `help logs` | 31 | syslog |
| `help nocharm` | 1 | medit-npc-flags (exact alias) |
| `help trigedit-commandlist` | 1 | trigedit-commands (exact alias) |
| `help gossip` | 1 | auctalk, chat, then auction |

The branch is squashed to one commit; the tree is byte-identical to `2b494c099`, which is what the local CI replica ran.

## Local CI replica on the final tree, all 33 matrix jobs

All 33 pass, including `test-unit-tests`, every `test-cmake-*` and `test-warning-budget-*` entry, both sanitizer jobs, `test-memory-check`, `test-fuzz`, `quality-clang-tidy`, `quality-format-check`, and `hygiene-hygiene`. Host run: `make test-all` against a prepared CI runtime with the database tests enabled is 1679/1679. Branch is 0 behind master, so no rebase was needed.

moshehbenavraham:
## Merged

Rebase-merged onto master as a single commit. The branch was squashed before the merge; GitHub rewrote the SHA:

| PR head | master |
| -- | -- |
| 72a972fa8 | a9bd7add1 |

`git diff 72a972fa8 origin/master` is empty.

**What landed**

- All three help lookup queries (`search_help`, soundex, edit distance) share `HELP_LOOKUP_KEYWORDS_SQL`, a recursive CTE that turns `help_entries.alternate_keywords` into lookup tokens alongside `help_keywords`. The 1,449 alias tokens from `help.hlp` now resolve in game. No schema change, no data migration, nothing to run on deploy.
- Ranking: exact token first, then keyword owner over alias, then shorter keyword. `help logs` shows SYSLOG, `help spell-resist` reaches MAGIC-RESISTANCE, `help gossip` keeps CHAT and AUCTALK ahead of AUCTION.
- The displayed keyword line includes aliases, so an alias hit is an exact match rather than a fall-through to suggestions.
- `Test_help_alias_tokens_are_lookup_keywords` covers the lookup, a prefix-keyword neighbor, the level gate, the collision order, and both suggestion queries on TEMPORARY shadow tables.
- `docs/systems/HELP_SYSTEM.md` and the production help log triage are updated.

**Follow-up**

- Misfiled aliases from the old flat-file import still route to the entry that carries them. That is a help data cleanup, out of scope here as the issue's caveat asked.
- #206 (`strip_colors()` overflow reached from the world loader) is unrelated to this diff and remains open.

Closes #204.

PR #207, opened 2026-09-19
Stop strip_colors at a trailing color marker

### LINKED ISSUE #206: strip_colors reads past the buffer on a trailing @ or tab, overflowing check_object's stack buffer during world load
## Problem

`strip_colors()` in `src/core/modify.c` reads past the end of its buffer when the string ends with a lone `@` (or a lone tab): on `@` followed by any byte other than `@` it does `p += 2`, so when that byte is the terminator the loop continues from one past the end.

`check_object()` in `src/core/db.c` copies each object's short description into a 512-byte stack buffer and calls `strip_colors()` on it, so an object file whose short description ends in `@` overflows that buffer during world load.

## Evidence

Found by the local replica of the `test-fuzz` job (fuzz target `world`, `fuzz_world_run()` in `unittests/CuTest/test_fuzz_targets.c`) on 2026-09-18 while validating PR #205, whose diff does not touch either function:

```text
ERROR: AddressSanitizer: stack-buffer-overflow
READ of size 1
    #0 strip_colors src/core/modify.c:105:10
    #1 check_object src/core/db.c:8066:3
    #2 parse_object src/core/db.c:4302:7
    #3 discrete_load src/core/db.c:2106:23
    #4 fuzz_world_run unittests/CuTest/test_fuzz_targets.c
Address is located in stack of thread T0 at offset 1216 in frame parse_object
    [704, 1216) 'buf1.i' (line 8059) <== Memory access at offset 1216 overflows this variable
```

## Where

- `src/core/modify.c` `strip_colors()`: the `@` and `\t` branches advance by two without checking for the terminator.
- `src/core/db.c` `check_object()`: 512-byte `buf1` fed to `strip_colors()`.

## Fix shape

Stop when the byte after `@` or `\t` is `\0` (drop the trailing marker), and add a CuTest that calls `strip_colors()` on `"name@"` and `"name\t"`.

moshehbenavraham:
## Summary

`strip_colors()` in `src/core/modify.c` skipped two bytes after a lone `@` or tab without checking for the terminator, so a string ending in either marker sent the scan past the end of the buffer. `check_object()` feeds every object's short description through it in a 512-byte stack buffer during world load, which the `world` fuzz target reported as an AddressSanitizer stack-buffer-overflow while validating #205.

- Stop at a trailing `@` or tab and drop the marker. Escaped markers (`@@`, tab tab) and complete codes are unchanged.
- Add `Test_strip_colors_stops_at_a_trailing_marker` to `unittests/CuTest/test_bounds_checking.c`. The helper pads the bytes after the terminator, so the old over-read changed the output and the test failed on the previous code without a sanitizer.

Closes #206

## Verification

- `CUTEST_FILTER=strip_colors ./cutest` fails on the old `strip_colors()` and passes with the fix.
- Local replica of `quality-clang-tidy`: baseline respected.
- Full local CI matrix (`scripts/ci/local/run.py --jobs 4 --cpus 4`): see the comment below.


moshehbenavraham:
Local CI on 4a776d93c (`scripts/ci/local/run.py --jobs 4 --cpus 4`, base 39f1f8d29): all 33 jobs passed in 599 s, including `test-unit-tests` (1680 tests), the three sanitizer jobs, `test-memory-check`, and `test-fuzz` (the `world` target that reported the overflow ran clean with and without the regression corpus). `quality-clang-tidy` ran alone beforehand: baseline respected.

The new test fails on the previous `strip_colors()` (`nameXY` instead of `name`) and passes with the fix. Ready for review.

moshehbenavraham (commented):
## Adversarial review of `4a776d93c` (issue 206)

Hunted the changed `strip_colors` skip/terminator paths, `Test_strip_colors_stops_at_a_trailing_marker`, and every production `strip_colors(` caller that feeds it world, player, script, or MSDP strings. Stance was “bugs exist.” Nothing reproduced on this HEAD as a correctness, security (OOB read/write), race, or stored-data-loss defect. Empty finding set — not an approval.

### Surfaces checked

- `src/core/modify.c` `strip_colors` (the `*(p + 1) == '\0'` `break` on both `@` and tab).
- CuTest helper + `Test_strip_colors_stops_at_a_trailing_marker` (padded `X` after the terminator so a leftover over-read changes the string without ASan).
- `src/core/db.c` `check_object` stack `buf1` (the ASan frame from #206), zone-name boot log copy, `save_objects_to_database` catalog copy.
- `src/dgscript/dg_triggers.c` `act_mtrigger` in-place `str`/`arg`.
- Other live callers: `src/core/comm.c` MSDP (`set_msdp_plain_text`, room table, title/pos/race/class, 255-byte opponent/tank name), `src/act/act.informative.c` 200-byte `obj_desc`, `src/act/act.other.c` group/inventory MSDP, `src/act/act.wizard.c`, `src/craft/craft.c` / `crafting_new.c` (heap `obj->name` after `strdup`), `src/character/vampire_cloak.c` / `roleplay.c`, `src/quest/quest.c`, `src/wilderness/terrain_bridge.c` stack copies, `src/net/i3_client.c`.
- Python mirror `scripts/world/wtool_lib/semantics.py` `strip_color_codes` only for world-data divergence.

### What was driven on this HEAD (not a reimplementation)

- `make cutest` then `CUTEST_FILTER=strip_colors ./cutest` → `OK (1 test)`.
- gdb `call strip_colors` on the symbol in that `./cutest` binary:
  - `"name@"` / `"name\\t"` → `"name"` (tab case bytes `6e 61 6d 65 00`).
  - `"@@"` → `"@"`, `"@@@"` / `"hello@@@"` → `"@" / `"hello@"` (escaped pair kept, leftover marker dropped).
  - mixed `"a@Rb@@c\\tnd\\t\\te"` bytes `61 62 40 63 64 09 65 00`.
  - **check_object replica:** 512-byte buffer, `'@'` at index 510, NUL at 511 → `strlen == 510`, byte 509 still `'A'`. Same for a 255-byte opponent-name replica (`len == 253`).
- ASan (`abort_on_error=1`) on the verbatim function body from `modify.c:98-142` with those same buffers, heap `strdup("name@")`, and NULL: clean.

Old `p += 2` on that 512-byte case is a 1-byte read at index 512 (the #206 ASan report). After the `break`, `*n = '\\0'` lands on the marker, inside the allocation. `n` only advances on kept bytes so it cannot write past the original terminator.

### Discarded (not posted as findings)

- **`act_mtrigger` in-place strip + `strchr(..., '\\r')`.** Zero call sites in the tree. Dead. `@\\r` is still a two-byte skip (not a trailing NUL marker); pre-existing if the function is ever revived.
- **Python vs C.** `strip_color_codes` already dropped a trailing marker by advancing `index` past `len`. Matches C on the cases above. No world-data mis-validate from this PR.
- **`mysql_real_escape_string` into `object_name[255]` needing `2*len+1`.** Pre-existing, untouched. A trailing `@` with no SQL-specials now stays inside that 255-byte buffer instead of skipping past it.
- **Two-byte skip vs `@[bold]` / long `\\t[` sequences.** Pre-existing parser shape; not an OOB and not this diff.
- **Truncation splitting `@@` on a strncpy/snprintf copy** so a log/MSDP/display buffer drops a leftover `@`. Copy only; world/proto storage is not written back. The alternative is the overflow this PR closes.
- **craft/cloak keyword `strip_colors` on `strdup`.** Dropping an incomplete trailing marker is the color-strip contract (`@@` still yields a literal `@`). Previously this was a heap over-read/write on an exact-size allocation.
- **Races.** Single-threaded loop; live callers copy to stack first or own the `strdup`.

No REQUEST_CHANGES items. Not approving.

